PR #56, opened 2015-02-12
Implement character trimming

astamm:
Main:
* Trimming was optional in character_parser() but actually not implemented.

Additionally:
* change integer parsing from 0 to -1 for unrecognised integers;
* count lines correctly if last line of file has no '\n' line break, which happens to some CSV files.

### DIFF
--- a/R/parsers.R
+++ b/R/parsers.R
@@ -19,7 +19,7 @@ NULL
 #' @param trim If \code{TRUE}, will trim off any leading or trailing whitespace.
 #' @rdname parser
 #' @export
-character_parser <- function(trim = FALSE) {
+character_parser <- function(trim = TRUE) {
   parser("character", trim = trim)
 }
 
--- a/R/parsers.R
+++ b/R/parsers.R
@@ -8,7 +8,7 @@ print.parser <- function(x, ...) {
 }
 
 find_parser <- function(name) {
-  match.fun(paste0(name, "_parser"))()
+  get(paste0(name, "_parser"), envir = asNamespace("fastread"))()
 }
 
 #' Define how to parse each column
--- a/inst/include/fastread/Source.h
+++ b/inst/include/fastread/Source.h
@@ -43,7 +43,7 @@ namespace fastread {
             return res ;
         }
 
-        SEXP get_String(){
+        SEXP get_String(bool trim = true){
             char* q = p ;
             int len = move_until_next_token_start() ;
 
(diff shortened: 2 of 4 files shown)


hadley:
Thanks! Would you mind adopting our C++ style (which isn't super consistent, but we don't use a new line for `{`)

astamm:
* I uniformized my C++ chunk code to match your style. 
* I changed "match.fun" to "get" in find_parsers because "get" has the option to specify the environment while match.fun does not. With match.fun, we had to load library(fastread) to enable auto detection of column parsers. Now we do not need to do that. 
* I thought about this auto detection feature because it failed me in one of my projects. Currently, it loads the first 30 lines out of which it guesses column types.
  * I have a variable that enumerates the different features present in an image. Features are coding from 1 to 18. Most of the time, images only contain 1 of the features but some of them towards the end of the file (~200 lines) have more features. Hence, I have something like sample(1:18, n = 100), "1,4,8"\n for example. In this case, fastread reads it as integer and parse the "1,4,8" into 3 columns instead of keeping them all into the same cell, which results in a major confusion for the whole dataset, for which not even a warning is displayed.
  * This made me think about auto detection and to me the only way to be sure that the "guess" is right is to guess according to all the lines in the dataset, which is easily changeable in the code. However, that means reading twice the data: one time for guessing types and another time for actually importing data, which is a waste of time. I think of these two strategies that might be better:
    * Leave the strategy as it is but instead of picking the first 30 lines, pick 30 random lines. Or come up with a randomized strategy that makes the guessing almost always right while keeping the number of lines read for guessing low; or, 
    * Systematically guess from all lines but keep the data imported the first time and format it according to guessed type, to avoid reading the data again.

Let me know your thoughts on this. I will also propose an implementation for dealing with NA soon. To my understanding, for now, you are only using the na_strings argument for guessing column types but they are not dealt with during import yet.

hadley:
I'd rather stick with just the first 30 lines - the goal is not to be perfect, but instead to have make a reasonable first guess that you can then modify if it's not correct.

astamm:
Understood. In this case, it would probably make sense to output a warning. As of now, data import went completely successful in my example although the imported data was completely off.

Although this situation is manageable as is for a ~100 line dataset, with bigger datasets, it would be nice to be able to assume that if read_delim does not display errors or warnings then the import went truly well.

On Feb 15, 2015, at 7:41 PM, Hadley Wickham <notifications@github.com<mailto:notifications@github.com>> wrote:


I'd rather stick with just the first 30 lines - the goal is not to be perfect, but instead to have make a reasonable first guess that you can then modify if it's not correct.

hadley:
Yes, absolutely. fastread is still very much a work in progress - we'll be putting in a lot more time on it in the next couple of months.

astamm:
I really like it and whenever I find time I am happy to contribute. (almost ready to share a first solution for handling NAs). I guess you are busy with finishing the writing of your book soon to be published :-)

Will try to help on fastread as I really like its philosophy (and speed!).

On Feb 15, 2015, at 7:48 PM, Hadley Wickham <notifications@github.com<mailto:notifications@github.com>> wrote:


Yes, absolutely. fastread is still very much a work in progress - we'll be putting in a lot more time on it in the next couple of months.

PR #57, opened 2015-02-18
Doc tweaks

Ironholds:
"inline" and "in line" are different things. Also, expand on the 'delim' description to give some examples (I expect anyone who uses fastread to know what a delimiter is, but I pray for a day when fastread or something like it is just SOP and something new users are exposed to. In the meantime, example-driven docs cost nothing).

### DIFF
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -8,7 +8,7 @@
 #'
 #' @param file Path to a file, or a connection object.
 #' @param delim the field delimiter to use, such as "," for a comma-separated file,
-#' or "\t" for a tab-separated file.
+#'   or "\t" for a tab-separated file.
 #' @inheritParams column_spec
 #' @param ... Other parameters passed on to \code{\link{build_delim_spec}}.
 #'   These are not normally necessary, but maybe be useful if you have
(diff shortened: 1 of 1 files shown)


R/read_delim.R
```diff
@@ -2,18 +2,19 @@
...
 #' provides substantially more feedback about what's going on. It produces a
-#' data frame inline with modern R usage: col names are left as is, factors
+#' data frame in line with modern R usage: col names are left as is, factors
 #' are never created by default, and row names are never used. \code{read_csv}
 #' is a version of \code{read_delim} where \code{delim} is set to ",".
 #'
 #' @param file Path to a file, or a connection object.
-#' @param delim Field delimiter.
+#' @param delim the field delimiter to use, such as "," for a comma-separated file,
+#' or "\t" for a tab-separated file.
```
hadley: Can you please indent the second line here?

Ironholds: What do you mean? It looks indented to me. DO you mean offset from the @param tag?

hadley: Yes - see the style for `@param ...` - I think this makes it easier to scan the docs

Ironholds: Agreed, and fixed!

Issue #64, opened 2015-03-10
The 'NA' parameter does not change NAs


Ironholds:
Using the test data in the testthat directory:

> test_data <- read_csv(file.path(getwd(),"/basic-df.csv"), na = "turnip")
> test_data
       a  b          c  d
1  FALSE  7 0.18152664  m
2  FALSE  2 0.83322744  z
3  FALSE  8 0.92679048  r
4  FALSE 10 0.37527031  s
5  FALSE  6 0.42026629  g
6  FALSE  3 0.43544999  h
7  FALSE  5 0.02109420  w
8  FALSE  9 0.09155708  u
9  FALSE  1 0.75610687  l
10 FALSE  4 0.35353098 NA

astamm:
I have worked on that on my fork.

Still have to make a pull request and check with Hadley & Romain if it fits the way they thought to handle NAs.

Will make the pull request tmrw.

-Aymeric

On Mar 9, 2015, at 9:04 PM, Oliver Keyes <notifications@github.com<mailto:notifications@github.com>> wrote:


Using the test data in the testthat directory:

test_data <- read_csv(file.path(getwd(),"/basic-df.csv"), na = "turnip")
test_data
a b c d
1 FALSE 7 0.18152664 m
2 FALSE 2 0.83322744 z
3 FALSE 8 0.92679048 r
4 FALSE 10 0.37527031 s
5 FALSE 6 0.42026629 g
6 FALSE 3 0.43544999 h
7 FALSE 5 0.02109420 w
8 FALSE 9 0.09155708 u
9 FALSE 1 0.75610687 l
10 FALSE 4 0.35353098 NA

Ironholds:
Cool!

hadley:
@astamm probably since you started working on that I completely rewrote the backend :/

hadley:
There was a bug, but I don't think it works the way you expect. Hopefully this makes it more clear:

```R
parse_vector("1,2,3,.", tokenizer_csv(na = "."), col_double())
parse_vector("1,2,3,NA", tokenizer_csv(na = "NA"), col_double())
```

hadley:
@astamm if you're interested in contributing more pull requests, I've written up some basic design docs in `vignettes/design.Rmd`. The new design should hopefully be easier to work with - please let me know if anything isn't clear.

astamm:
@hadley yes I would like to contribute to the package as much as possible.

Will read the design vignette soon.

Thanks!
[...]

Issue #63, opened 2015-03-10
Using n_max introduces corrupt records


Ironholds:
Using the test data in the testthat directory:

>test_data <- read_csv(file.path(getwd(),"/basic-df.csv"), n_max = 7)
>test_data

>   a  b         c    d
1 FALSE  7 0.1815266    m
2 FALSE  2 0.8332274    z
3 FALSE  8 0.9267905    r
4 FALSE 10 0.3752703    s
5 FALSE  6 0.4202663    g
6 FALSE  3 0.4354500    h
7 FALSE NA        NA NA

PR #65, opened 2015-03-10
New unit tests

Ironholds:
Looking at the test dataset in the testthat directory, I imputed (yay, R docs teaching me new words) that you wanted to test the file reading.

This patch:

1. Introduces a set of unit tests for read_csv. It's not all-inclusive but it's a start. It covers a basic test that each of the parameters actually works and doesn't corrupt the file.
2. Modifies the basic-df.csv file to include an NA value, thus enabling the NA parameter in read_csv to be tested (this turned out to be important. See below)
3. Introduces _specific_ tests for bugs I discovered while writing the tests, which should protect against regression in the long-term.

While writing tests I discovered two bugs in CSV reading, which are reported and documented as #63 and #64 respectively.


Ironholds:
(It goes without saying that as a result of "writing unit tests for things that were specifically bugs" and "discovering bugs when my unit tests blew up", this patch means readr will not make a clean sweep of the unit tests until the bugs are fixed, which I'm happy to distinctly work on if you want to assign me one of them)

tests/testthat/test-read-csv.R
```diff
@@ -0,0 +1,37 @@
+context("read_csv parameter testing")
+
+#Formatting for column imputation, col_name detection and NA detection.
+test_data <- read_csv(file.path(getwd(),"/basic-df.csv"), col_types = NULL, col_names = TRUE)
+
+test_that("read_csv automatic column type imputation works",{
+  expect_that(unname(unlist(lapply(test_data,class))), equals(c("logical","integer","numeric","character")))
```
hadley: I prefer to use `expect_equals()` style. Would you mind updating please?

tests/testthat/test-read-csv.R
```diff
@@ -0,0 +1,37 @@
...
+})
+
+test_that("read_csv's col_names = TRUE actually picks up col_names",{
+  expect_that(names(test_data), equals(c("a","b","c","d")))
```
hadley: If you're using the same parsed file, I think it's appropriate to put all the expectations in one test

tests/testthat/test-read-csv.R
```diff
@@ -0,0 +1,37 @@
...
+})
+
+test_that("read_csv detects NA values",{
+  expect_that(sum(is.na(test_data$d)), equals(1))
+})
+
+test_that("read_csv's 'NA' option genuinely changes the NA values",{
+  test_data <- read_csv(file.path(getwd(),"/basic-df.csv"), na = "turnip")
```
hadley: Just `"basic-df.csv"` should be fine - tests are run with wd set to `tests/testthat`

tests/testthat/test-read-csv.R
```diff
@@ -0,0 +1,37 @@
...
+  expect_that(nrow(test_data[test_data$d == "turnip",]), equals(1))
```
hadley: But I think you're misinterpreting the value of `na` - it's what input strings will be recognised as NA, not how NAs will be represented in the output

tests/testthat/test-read-csv.R
```diff
@@ -0,0 +1,37 @@
...
+})
+
+test_that("read_csv's 'skip' option allows for skipping'",{
+  test_data <- read_csv(file.path(getwd(),"/basic-df.csv"), skip = 1)
```
hadley: It would be useful to test the interaction of `skip` and `col_names` - if `col_names` is TRUE, it should skip the first two rows

tests/testthat/test-read-csv.R
Ironholds: *headdesks*. My bad. I'll (distinctly) throw in a patch clarifying the documentation there.

(As an aside: it appears to not work even factoring that in. Aw :()

tests/testthat/test-read-csv.R
Ironholds: Makes sense!

Ironholds:
@hadley issues should be resolved!

tests/testthat/test-read-csv.R
robertzk: At least on the Ruby side, I hear general TDD practice is to put one expectation per test, so that test failures narrow down to a single assertion and minimize further debugging. What do you think of

```r
describe("using read_csv on basic-df.csv", {
   # code from above
})

hadley:
Thanks!

hadley:
Had to make a couple of small tweaks - I think the two failures were bad tests

Ironholds:
Bah! My apologies :(

PR #83, opened 2015-03-13
Fix: col_euro_double() should be called col_eurodouble()

juba:
Trying to load a simple CSV file including a column with comma as decimal separator fails with the following message : 

    Error in get(paste0("col_", name), envir = asNamespace("readr")) : 
      object 'col_eurodouble' not found

Renaming `col_euro_double()` into `col_eurodouble()` in `collectors.R` seems to fix the problem.


hadley:
Seems a bit simpler to fix in the C++

PR #86, opened 2015-03-13
Typo in read_csv2

dickoa:
In `read_csv2`, I think `delim` parameters should be `;` instead of `.`.
Thanks again for this wonderful package


hadley:
Oh oops, thanks

PR #101, opened 2015-03-25
Give the appropriate warning for integers.

johnmcdonnell:
I think this should cause the proper warning when integers fail to parse.

With the usual caveats:

I have no idea what I'm doing


johnmcdonnell:
I'm not super familiar with this build process, but I don't think this commit broke the build.

     ERROR: configuration failed for package ‘V8’

hadley:
Thanks! I'm pretty sure the break is unrelated to your change

PR #102, opened 2015-03-25
By default, scan through whole file to determine types.

johnmcdonnell:
This is the nuclear method, it does have considerable performance impacts. I'm not completely sure how else to be completely certain of getting the types right besides scanning the whole file.

I'm still getting much better performance than ``read.csv``:

    > system.time(x <- read_csv('parse_resistant_data.csv'))
    user  system elapsed 
    4.312   0.281   5.345 
    > system.time(y <- fread('parse_resistant_data.csv'))
       user  system elapsed
      0.621   0.054   0.675
    > system.time(z <- read.csv('parse_resistant_data.csv'))
       user  system elapsed
     16.484   0.366  17.973

A really crazy alternative might be to watch for parse warnings when reading the whole file, and just start over with a full file scan if they do happen. That would result in unpredictable performance but it would be fast in most cases where the types are picked up correctly on the first run.

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -33,7 +33,7 @@ read_tokens <- function(sourceSpec, tokenizerSpec, colSpecs, col_names, n_max = 
     .Call('readr_read_tokens', PACKAGE = 'readr', sourceSpec, tokenizerSpec, colSpecs, col_names, n_max, progress)
 }
 
-collectorsGuess <- function(sourceSpec, tokenizerSpec, n = 100L) {
+collectorsGuess <- function(sourceSpec, tokenizerSpec, n = -1L) {
     .Call('readr_collectorsGuess', PACKAGE = 'readr', sourceSpec, tokenizerSpec, n)
 }
 
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -29,8 +29,8 @@ NULL
 #' # Or directly from a string (must contain a newline)
 #' read_csv("x,y\n1,2\n3,4")
 #'
-#' # By default, readr guess the columns types, looking at the first 100 rows.
-#' # You can override with a compact specification:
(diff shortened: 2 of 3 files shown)


johnmcdonnell:
It looks like `fread` uses a fairly janky guessing method:

    Make best guess at column types using first 5 rows, middle 5 rows and last 5 rows

https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L822

johnmcdonnell:
From fread, on line 365:

> warning("Bumped column %d to type character on data row %d, field contains '%.*s'. Coercing previously read values in this column from logical, integer or numeric back to character which may not be lossless; e.g., if '00' and '000' occurred before they will now be just '0', and there may be inconsistencies with treatment of ',,' and ',NA,' too (if they occurred in this column before the bump). If this matters please rerun and set 'colClasses' to 'character' for this column. Please note that column type detection uses the first 5 rows, the middle 5 rows and the last 5 rows, so hopefully this message should be very rare. If reporting to datatable-help, please rerun and include the output from verbose=TRUE.\n", col+1, sofar+1, lch-ch, ch);

...so sounds like fread uses slightly different behavior, but doesn't make an attempt to re-derive the type of the column. I somewhat prefer the char coerce method to misparsing to NA because I think a) the user is more likely to realize what has happened and b) char will often be a very recoverable outcome (e.g., floats coerced to chars can be cast back to floats). Obviously there isn't really an ideal solution that preserves performance.

hadley:
I think readr is going to have a different philosophy to that - if its simple heuristics fail, you'll need to override the column types yourself. This needs better documentation, and I have some ideas on how to display the warnings so that (a) they're not so slow and (b) you can see more of them in a convenient format.

johnmcdonnell:
Makes sense, speed and warning transparency would greatly improve usability in these cases.

PR #104, opened 2015-04-01
Add read_markdown_table

sjackman:
Read Markdown tables. For example:

```
A|B|C|D
-|-|-|-
0|1|2|3
4|5|6|7
```

| A | B | C | D |
| --- | --- | --- | --- |
| 0 | 1 | 2 | 3 |
| 4 | 5 | 6 | 7 |

To do:
- Determine automatically whether the column names are present.
- Remove leading and trailing white space around each cell.


hadley:
This doesn't seem quite right to me - you're manually dropping the second line (i.e. the dashes) but the column guesser will still see it so I think every column will be imported as text. Instead I think you'll need to read the column names yourself (since they're always required in this format), then call `read_delim()` with those names and `nskip = 2`. Does that make sense?

sjackman:
Hi, Hadley. The column names are in fact optional (see below), at least for Pandoc. The row of hyphens is not optional. See [Extension: pipe_tables](http://johnmacfarlane.net/pandoc/demo/example9/pandocs-markdown.html#extension-pipe_tables). GitHub Markdown preview doesn't like a missing header line. Good observation about the column formats. I hadn't noticed that. I've updated the PR.

It duplicates a lot of code from `read_delimited`. Would you rather add an additional argument to `read_delimited` named `skip_after_header`?

```
-|-|-|-
0|1|2|3
4|5|6|7
```

<table>
<tbody>
<tr class="odd">
<td align="left">0</td>
<td align="left">1</td>
<td align="left">2</td>
<td align="left">3</td>
</tr>
<tr class="even">
<td align="left">4</td>
<td align="left">5</td>
<td align="left">6</td>
<td align="left">7</td>
</tr>
</tbody>
</table>

sjackman:
@hadley I'd love for this function to make it into your CRAN submission. I'm happy to do what's needed to get it in. How are you feeling about this PR? Would you prefer adding the option `skip_after_header` to `read_delimited` to avoid repetition of code?

hadley:
I think this looks ok, but I'd rather call it `read_md_table()` (or maybe `read_table_md()`?), and it needs a few unit tests.

hadley:
If you can do by Friday, I can include in the next version of readr

sjackman:
Hi, Hadley. Yep, I think I can get this done by Friday. I'll need to add a `skip_after_header` parameter to both `col_spec_standardise` and `read_delimited` to skip the row of hyphens following the header line. Does that work for you?

hadley:
Hmmm, I really want to avoid have skip before and after header parameters.

sjackman:
I can duplicate the code of `read_delimited`—I've already done so in this PR—but I'd rather not duplicate the code of `col_spec_standardise`. Any alternative suggestion?

sjackman:
I could see a `skip_after_header` option also having a use in loading the last million records of a billion-record TSV file.

hadley:
Hmmm, I think that calls for a more generalised solution which I don't have time to think about at the moment. Lets try again for the next version.

jimhester:
While this idea I think is interesting in practice the number of documents which contain _just_ markdown tables and nothing else is very small, so it is hard to envision cases when this would be used.

Since it has been languishing for a number of years I think it should be closed.

sjackman:
No worries at all, Jim.

We discussed reading Markdown tables using `rio` over at https://github.com/leeper/rio/issues/70
The final solution was to convert Markdown tables to HTML, and then read the HTML tables, which `rio` already supports.
As a bonus, `rio` supports reading multiple tables from a single document https://github.com/leeper/rio/issues/126

PR #123, opened 2015-04-12
Update README.md

kbenoit:
updated to reflect the fact that readr is now on CRAN. Keep up the excellent work!


hadley:
Would you mind including the actual install command for the CRAN package (i.e. like https://github.com/hadley/rvest)?

kbenoit:
Done.

hadley:
Thanks!

Issue #135, opened 2015-04-16
Implement write_tsv


hadley:
Should be simple generalisation of `write_csv`. Use same strategy to only quote when needed.

nicolasCoutin:
@hadley I've been reading through your [r-pkgs book](http://r-pkgs.had.co.nz/) book and I'm learning me how to package. It's very helpful! Anyway since I wanted to walkthrough the process with a simple use case:  

 i decided to make [`writer`](https://github.com/nicolasCoutin/writer) where I (stole! and) generalised `readr:::stream_csv()` to `stream_delim()` and used it to implement:   

-  `write_csv()` 
- `write_tsv()`  
- `write_delim()`  

 I also threw in naive wrappers around `base::saveRDS` to make:
- `write_rds()` 
- `write_compressed_rds()`  

Now although this was purely an exercise to get learning.. it got me wondering: If `readr` is about reading then should you have a `writer` (or maybe `saver` since it has 5 letters) package? And if `readr` is really about stable/streaming data input and output should it then be renamed to reflect this? I guess I'm really asking about the phylosophy of how package-happy one should be when developing thingamabobs.

I hope it's helpful with this issue!
If you have any thoughts `rep(c('thanks', '!'),1E6`)

hadley:
I'd rather just keep all the functionality in readr - I don't think it's too awful to keep the name as readr, since that is the primary purpose. Do you want to do a pull request?

nicolasCoutin:
@hadley okee: see #157

Issue #140, opened 2015-04-17
Add read_rds/write_rds aliases


hadley:
(possibly with different compression defaults to increase speed)

hadley:
Seems reasonable to default to `compress = FALSE`, and document the tradeoffs.

PR #157, opened 2015-04-30
Implement `write_tsv()`, `write_delim()`, `write_rds()`

nicolasCoutin:
Also see #135 .  

- Generalised `stream_csv()` to `stream_delim()` to have as common worker for `write_csv()`, `write_tsv()` and `write_delim()`.

- light wrappers around `base::saveRDS()` gives `write_rds()` which never compresses for speed and `write_compressed_rds()`. The latter currently defaults to use `xzfile(..., compression = 9L)` as this generates the smallest file for most cases in my experience. Could try to implement a `guess_compression()` which picks smart defaults based on object size and type in the future.


R/write.R
```diff
@@ -32,11 +52,29 @@ write_csv <- function(x, path, append = FALSE, col_names = !append) {
...
-#' @rdname write_csv
+#' @rdname write_delim
+#' @export
+#' @examples
+#' cat(write_tsv(head(mtcars), ""))
+#' cat(write_tsv(head(iris), ""))
+#'
+#' df <- data.frame(x = c("a", '"', ",", "\n"))
+#' read_tsv(write_tsv(df, ""))
+write_tsv <- function(x, path, append = FALSE, col_names = !append) {
```
hadley: Could this and `write_csv()` just call `write_delim()`?

R/write.R
```diff
@@ -58,3 +96,42 @@ output_column.POSIXt <- function(x) {
...
+#' ## write_rds(mtcars, "mtcars.rds")
+write_rds <- function(x, path) {
+  saveRDS(x, path, ascii = FALSE, version = NULL, compress = FALSE, refhook = NULL)
+}
+
+#' @rdname write_rds
+#' @export
+#' @examples
+#' ## write_compressed_rds(mtcars, "mtcars.rds", compress_fun = gzfile)
+write_compressed_rds <- function(x, path, compress_fun = xzfile, compress_amount = 9L, ...) {
```
hadley: I think I'd prefer this to just just one function, where `compress` defaults to FALSE

hadley: Actually might be better to have `compress = c("none", "gz", "xz", "bz"))` or similar

src/write_delim.cpp
```diff
@@ -28,7 +28,7 @@ bool needs_quote(const char* string, char delim = ',') {
 }
 
 template <class Stream>
-void stream_csv(Stream& output, const char* string) {
+void stream_delim(Stream& output, const char* string) {
   bool quotes = needs_quote(string);
```
hadley: I think `needs_quote` now needs to get parsed the delimiter

hadley:
A couple of unit tests for tsv and delim would be useful to make sure that quotes get added in the right places.

nicolasCoutin:
@hadley I think gotten at each of your comments:

- `write_csv()` and `write_tsv()` are now just wrappers around `write_delim()` with `delim = ","` and `delim = "\t"` respectively.  
- modified `need_quote()` in write_delim.cpp to _always_ spot `,` as well as passing in `delim`.  
- `write_rds()` is now the only function with the expected arguments and defaults.  
- added tests for `write_tsv()` and `write_delim()`...see below...


N.B. writing tests are brand new to me but they seem to point out something interesting in `stream_delim()`'s handling of in-field `"` ...which also was happening in `stream_csv()`:

I.e. the current (`devtools::install_github(hadley/reader@master)`):
```r
x <- c("a", '"', ",", "\n", "at\t")
output <- data.frame(x)
readr:::stream_csv(lapply(output,output_column), '')
#> [1] "x\na\n\"\"\"\"\n\",\"\n\"\n\"\nat\t\n"
```
I am just be sleep deprived but this is a bug right? --> refer to test-write-delim.R which catches this... Also a general testing question: In a case like this it seems like it would be good to test the unexported `stream_delim()` directly (as in on top of). I did see (as you wrote) :

> Focus on testing the external interface to your functions - if you test the internal interface, then it’s harder to change the implementation in the future because as well as modifying the code, you’ll also need to update all the tests.

R/write.R
```diff
@@ -58,3 +84,34 @@ output_column.POSIXt <- function(x) {
...
+NULL
+
+#' @rdname write_rds
+#' @export
+#' @examples
+#' ## write_rds(mtcars, "mtcars.rds")
+#' ## write_rds(mtcars, "compressed_mtc.rds, "xz", compression = 9L)
+write_rds <- function(x, path, compress = c("none", "gz", "bz", "xz"), ...) {
+
+  if(length(compress)!=1) compress <- compress[[1]]
```
hadley: I'd do 

```R
compress <- match.arg(compress)
switch(compress, 
  none = file, 
  ...
)
```

src/write_delim.cpp
```diff
@@ -4,23 +4,23 @@ using namespace Rcpp;
...
+      output << delim;
   }
   output << '\n';
 }
 
-bool needs_quote(const char* string, char delim = ',') {
+bool needs_quote(const char* string, char delim) {
   for (const char* cur = string; *cur != '\0'; ++cur) {
-    if (*cur == '\n' || *cur == '\r' || *cur == '"' || *cur == delim)
+    if (*cur == '\n' || *cur == '\r' || *cur == '"' || *cur == ',' || *cur == delim)
```
hadley: I think you can remove `*cur == ','`

tests/testthat/test-write-tsv.R
```diff
@@ -0,0 +1,36 @@
+context("write_tsv")
+
+test_that("strings are only quoted if needed", {
```
hadley: Given that `write_tsv()` just calls `write_delim()`, I don't think adding specific tests adds anything. I'd lump together all the writer functions in one set of tests - I think the main thing to check is the `write_delim()` correctly round trips regardless of the delimiter used.

hadley:
I'm not sure what you're seeing with the `stream_csv` test. This looks ok to me:

```R
x <- c("a", '"', ",", "\n", "at\t")
output <- data.frame(x, stringsAsFactors = FALSE)
cat(readr:::stream_csv(output, ''))
#> "x"
#> "a"
#> """"
#> ","
#> "
#> "
#> "at	"
```

nicolasCoutin:
@hadley What I was asking about was this: (again with `hadley/readr@master`)

```r
x <- c("a", '"', ",", "\n", "at\t")
output <- data.frame(x, stringsAsFactors = FALSE)

read_csv(readr:::stream_csv(output, ''))
#>      x
#> 1    a
#> 2    "
#> 3    ,
#> 4   \n
#> 5 at\t
read_delim(readr:::stream_csv(output, ''),delim = ',')
#>      x
#> 1    a
#> 2 """"
#> 3    ,
#> 4   \n
#> 5 at\t
```

I _thought_ they should both do a round trip but they don't...

hadley:
Oh hmmmm, that's a good point. Changing the default of `escape_double` seems reasonable to me.

nicolasCoutin:
This also means `escape_backlash` has to default to `FALSE` with current unescape method: 

https://github.com/hadley/readr/blob/e8728bd8d6f074e67797bd5c4716cb5e732ccff9/src/TokenizerDelim.h#L220-229

nicolasCoutin:
@hadley Ok:
-  `read_rds()` now uses `match.arg()` and `switch()` as advised.
- removed overkill always quote comma.
- thanks to changed `escape_double = TRUE` and `escape_backlash = FALSE` defaults in `read_delimited()` all round trips work as expected.  
- only one test file pertaining to `write_delim()` which covers for identical behaviour round trips w/ `read_csv()`, `read_tsv()`, and `read_delim()` given the respective `delim`.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # readr 0.1.0.9000
 
+  
+* Convenient wrapper around base::saveRDS `write_rds()` for saving a single R 
+  object (#140).
```
hadley: Can you please note your github username here? (e.g. `(#140, @nicolasCoutin)`)

R/read_delim.R
```diff
@@ -18,8 +18,8 @@ NULL
 #' @param progress Display a progress bar? By default it will only display
 #'   in an interactive session. The display is updated every 50,000 values
 #'   and will only display if estimated reading time is 5 seconds or more.
-#' @usage read_delim(file, delim, quote = '\"', escape_backslash = TRUE,
-#'   escape_double = FALSE, na = "NA", col_names = TRUE, col_types = NULL,
+#' @usage read_delim(file, delim, quote = '\"', escape_backslash = FALSE,
+#'   escape_double = TRUE, na = "NA", col_names = TRUE, col_types = NULL,
```
hadley: This also needs to be noted in NEWS

R/write.R
```diff
@@ -58,3 +84,36 @@ output_column.POSIXt <- function(x) {
   format(x, "%Y-%m-%dT%H:%M:%OSZ", tz = "UTC")
 }
 
+#' Write a single R object to file
+#'
+#' Consistent wrapper around \code{\link{saveRDS}}. \code{write_rds} never
+#' compresses by default as space is generally cheaper than time.
+#'
+#' @section guess compression:
```
hadley: Drop this section for now?

R/write.R
```diff
@@ -58,3 +84,36 @@ output_column.POSIXt <- function(x) {
...
+#' N.B. this is not yet implemented but some empirical testing may help come up
+#' with something useful.
+#' @param x R object to write to serialise.
+#' @param path Path to write to.
+#' @param compress Compression method to use one of \code{c("none", "gz" ,"bz",
+#' "xz")}.
+#' @param ... Additional arguments to connection function.
```
hadley: "For example, `compression_level`, allows you to control the space-time tradeoff for the compression methods)

R/write.R
```diff
@@ -58,3 +84,36 @@ output_column.POSIXt <- function(x) {
...
+
+#' @rdname write_rds
+#' @export
+#' @examples
+#' ## write_rds(mtcars, "mtcars.rds")
+#' ## write_rds(mtcars, "compressed_mtc.rds", "xz", compression = 9L)
+write_rds <- function(x, path, compress = c("none", "gz", "bz2", "xz"), ...) {
+
+  compress <- match.arg(compress)
+  switch(compress,
```
hadley: You can make this a bit DRYer by moving the assignment outside of the switch:

```R
con <- switch(compress,
  none = file(path, ...),
  ...
)
```

R/write.R
```diff
@@ -58,3 +84,36 @@ output_column.POSIXt <- function(x) {
...
+         none = con <-    file(path, ...),
+         gz   = con <-  gzfile(path, ...),
+         bz2  = con <-  bzfile(path, ...),
+         xz   = con <-  xzfile(path, ...))
+
+  saveRDS(x, con)
+  on.exit(close(con))
```
hadley: The `on.exit()` should go above `saveRDS()` in case `saveRDS` throws an error. It's also a good habit to always do `on.exit(..., add = TRUE)`, so if you have multiple `on.exit()`s they don't override each other

hadley:
A few more minor things. Getting very close!

nicolasCoutin:
@hadley happy to learn. Please have a looksie when convenient! I've gone after each of your comments updating NEWS and improving/cleaning up `write_rds()` to be safer with clearer docs.

hadley:
Looks good - thanks!

PR #159, opened 2015-05-02
don't stop() if there are unmatched column types

### LINKED ISSUE #145: Matching the extra columns specified in col_types
I would like to specify the column types using arguments `col_types` when I read the csv files. However, the column number could be different in my multiple csv files (Some files include extra column). 

When I specify extra column types and get an error messages:

    Error: The following named parsers don't match the column names: sow_date1

It will be better to let us to specify extra columns for flexibility (or give a warning message).

alyst:
fixes #145

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,3 +9,6 @@
 * When guessing field types, and there's no information to go on, use
   character instead of logical (#124, #128).
 
+* `read_XXX()` methods would not throw an error if `col_types` argument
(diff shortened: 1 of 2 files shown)


hadley:
Could you please add a bullet point to news (including thanking yourself)?

alyst:
@hadley done

hadley:
Thanks!

PR #160, opened 2015-05-02
Fix type_convert()

alyst:
The current behavior of `type_convert()` is not very practical:
 * it automatically converts all the character columns into the guessed type, if none explicitly specified
 * the current implementation of `col_types_full()` works only under the assumption that all columns have guessed type

The PR tries to fix the two issues. However, it only works when the conversion is from character to any other type, if e.g. one needs to convert integer into numeric, it would fail. This needs to be fixed in `type_convert_col()`.

### DIFF
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -77,10 +77,13 @@ col_types_full <- function(col_types, col_names, guessed_types) {
         paste0(bad_names, collapse = ", "), call. = FALSE)
     }
 
-    skip <- setdiff(col_names, names(col_types))
-    if (length(skip) > 0) {
+    # if no names given for guessed_types, assume it matches the col_names order
+    if ( is.null(names(guessed_types)) && length(guessed_types) == length(col_names) )
       names(guessed_types) <- col_names
-      col_types[skip] <- lapply(guessed_types[skip], collector_find)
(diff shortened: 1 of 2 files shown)


hadley:
I don't quite understand the problem - could you please explain in a little more detail?

alyst:
Imagine the workflow when one first reads the data with e.g. `read_tsv()` partially specifying the column types. Then he/she fixes the data and calls `type_convert()` to do additional conversions.

So, by the time `type_convert()` is called, some columns are already non-character (converted by `read_tsv()`) and therefore not guessed, but `col_types_full()` would nevertheless try to convert them and get their type from `guessed_types`. The first commit fixes that by actually checking what columns are in the `guessed_types` list.

The 2nd problem is that `type_convert()` would implicitly try to convert all character columns, even if the user has not requested that. It is fixed by adding `guess` parameter to `type_convert()` in the 2nd commit + fixing the `type_convert()` behaviour when the assumption `names(df) == union(names(col_types), names(guesses))` does not hold.

Still missing is the conversion from non-character types to the type specifies in `col_types`.

hadley:
Could you please file an issue with a minimal reprex? It seems to work ok for me.

PR #178, opened 2015-05-22
Fix a bug in DateTimeParser where compound formats would get parsed as

kmillar:
partial ISO8601 date time format.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,3 +17,5 @@
 * When guessing field types, and there's no information to go on, use
   character instead of logical (#124, #128).
 
+* Compound formats %D, %F, %R, %X, %T, %x are now parsed correctly, instead of
+  using the ISO8601 parser (#178, @kmillar)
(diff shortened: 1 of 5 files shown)


hadley:
Would you mind explaining how the patch fixes the problem?

kmillar:
Previously there were two overloads of parse in the code: parse(bool) and parse(const std::string&)

In e.g. 
   parse("%m/%d/%y");
you have a 'const char*'

const char* can be converted to bool, and can also be converted to std::string, so either version is possible.

The rules of C++ say that the bool conversion takes precedence, which isn't what you want here.

Generally it's a good idea to only overload functions when the semantics of the function is the same no matter which overload is chosen, otherwise things like this happen.

hadley:
Thanks - that's very useful. I really should rename `parse(bool)` to `parseISO8601()`.

Would you mind adding a bullet to NEWS describing the bug and including `(#178, @kmillar)`

kmillar:
Your rename suggestion is a much better approach.

hadley:
Thanks!

PR #183, opened 2015-05-30
clarify write_* docs (append param).

### LINKED ISSUE #181: write_*(..., append = TRUE) creates new file
if I try to `append` to a __non-existing file__ it creates the file (_silently_).

```r
does_not_exist <- "dne.tsv" # this file/connection does not exist in my working directory
readr::write_tsv(mtcars, does_not_exist, append = TRUE, col_names = TRUE)
```
from `?write_tsv`:
```r
append	  If FALSE, will create a new file. If TRUE, will append to an existing file.
```

I think three options here:

1. simply note the behaviour more precisely in the documentation (this was unexpected for me based on reading it).
2. message / warn user if file did not exist and was created. (my least favourite).
3. additional argument in `write_delim` to make the user declare if they want to _both append if the file exists and create if it does not.

if @hadley thinks this is a thing I can have a go at a PR if that would help.

npjc:
Modified documentation for the `append` param for the `write_*` functions to make it clear that if a file does not previously exist a new file is created.

This is to address / fix #181 .

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -16,8 +16,9 @@
 #' @param x A data frame to write to disk
 #' @param path Path to write to. If \code{""} will return the csv file as
 #'   a string.
-#' @param append If \code{FALSE}, will create a new file. If \code{TRUE},
-#'   will append to an existing file.
+#' @param append If \code{FALSE}, will overwrite existing file. If \code{TRUE},
+#'   will append to existing file. In both cases, if file does not exist a new
+#'   file is created.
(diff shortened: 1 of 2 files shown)


npjc:
@hadley PTAL when you get a chance. I'm not understanding why this is failing on travis. It passes locally.

npjc:
re-open synched with new upstream master.

npjc:
@hadley refreshed this to be up-to-date to upstream/master.

hadley:
Thanks!

Issue #125, opened 2015-04-13
Multiple strings for missing values


TonyLadson:
I'm working on files that use many different strings to specify missing values e.g. "" (blank), NA, na, nr, nd. read_csv seems to only allow a single string.

    read_csv("x,y\n1,2\n3,4\nna,NA", na = c('NA','na'))
    Error: expecting a single value

It would be great if I could include multiple strings in the specification of missing values.

Thank you for all your work on readr.  It is very fast.

rpruim:
I'll second this motion.

hadley:
(This is the flip side of #114)

jennybc:
:+1:

mdlincoln:
:+1: on this, and again on #114 :smile:

Issue #114, opened 2015-04-10
na="" not working for type 'c' columns


bearloga:
Trying to read the CSV from the [Planned Parenthood competition](http://www.drivendata.org/competitions/6/) (can be downloaded by signing up)
```
R> col_types <- 'icn...nc...c' # shortened for GitHub
R> train_data <- read_csv('train_values.csv',col_types=col_types,na="")
```
Checking for empty string values:
```
R> train_data[,1370:1379] %>% apply(2,. %>% { .=="" } %>% sum)
c_1368 c_1369 c_1370 c_1371 c_1372 c_1373 c_1374 c_1375 c_1376 c_1377 
 14639  14561  13925  14642      4  13830   8852  13623  14636  11071 
```
Checking for NA values:
```
R> train_data[,1370:1379] %>% apply(2,. %>% is.na %>% sum)
c_1368 c_1369 c_1370 c_1371 c_1372 c_1373 c_1374 c_1375 c_1376 c_1377 
     0      0      0      0      0      0      0      0      0      0 
```
Comparing **n**umeric and **c**haracter type columns:
```
R> train_data[,115:125] %>% apply(2,. %>% is.na %>% sum)
n_0112 n_0113 n_0114 n_0115 o_0116 o_0117 o_0118 o_0119 o_0120 o_0121 o_0122 
 14594  12785  13833  14494      0      0      0      0      0      0      0 

R> train_data[,115:125] %>% apply(2,. %>% { .=="" } %>% sum)
n_0112 n_0113 n_0114 n_0115 o_0116 o_0117 o_0118 o_0119 o_0120 o_0121 o_0122 
    NA     NA     NA     NA  14322  14643  14597  14557     39  14643  14635 
```
Is this intended behavior?

hadley:
Simple reproducible example:

```R
read_csv("x,y
,
a,b", na = "")
```

This appears to be by design - I was using "" to indicate that you didn't want any missing values.

mdlincoln:
If so, then how do you indicate that you *do* want NAs for missing values in your character type columns, rather than empty strings as seems to be the current default behavior:

```R
read_csv("x,y
,
a,b")
```

```
Source: local data frame [2 x 2]

  x y
1    
2 a b
```

hadley:
I should've been clear that's its a bad design, and I'll fix it ;)

bearloga:
Awesome! Thank you for fixing this!

PR #214, opened 2015-07-07
Allow multiple strings for missing values and correctly treat na = ""

andres-s:
Addresses #125 and #114.

I tried to follow the conventions of `utils::read.table`, namely `na = NULL` indicates no missing values while `na = ""` indicates that empty strings should be interpreted as missing values:

```r
example = "a,b
1,foo
,
"

f = textConnection(example)
read.csv(f, colClasses = c(a = "integer", b = "character"), na.strings = NULL)
#>    a   b
#> 1  1 foo
#> 2 NA    
close(f)

f = textConnection(example)
read.csv(f, colClasses = c(a = "integer", b = "character"), na.strings = "")
#>    a    b
#> 1  1  foo
#> 2 NA <NA>
close(f)
```

The implementation of `Token::flagNA` is a simple loop, I'm happy to do a performance comparison with less naive implementations if the behaviour is correct (although I suspect that the simple loop will win for small numbers of `na` strings).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,3 +17,5 @@
 * When guessing field types, and there's no information to go on, use
   character instead of logical (#124, #128).
 
+* Multiple strings can now be specified to indicate missing values (#125).
+  Specifying `na = ""` now works as expected with character columns (#114).
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,5 +17,7 @@
 * When guessing field types, and there's no information to go on, use
   character instead of logical (#124, #128).
 
-* Multiple strings can now be specified to indicate missing values (#125).
+* Multiple strings can now be specified to indicate missing values (#125), and
+  the default for missing strings has been changed to `na = c("", "NA")`.
   Specifying `na = ""` now works as expected with character columns (#114).
+  
--- a/R/tokenizer.R
+++ b/R/tokenizer.R
@@ -35,7 +35,7 @@ NULL
 #' @export
 #' @rdname Tokenizers
 #' @param na Character vector of strings to use for missing values. Set this
-#'   option to \code{NULL} to indicate no missing values.
+#'   option to \code{character()} to indicate no missing values.
 #' @param delim Single character used to separate fields within a record.
 #' @param quote Single character used to quote strings.
 #' @param escape_double Does the file escape quotes by doubling them?
--- a/R/tokenizer.R
+++ b/R/tokenizer.R
@@ -34,7 +34,8 @@ NULL
 
 #' @export
 #' @rdname Tokenizers
-#' @param na String to use for missing values.
+#' @param na Character vector of strings to use for missing values. Set this
+#'   option to \code{NULL} to indicate no missing values.
 #' @param delim Single character used to separate fields within a record.
 #' @param quote Single character used to quote strings.
 #' @param escape_double Does the file escape quotes by doubling them?
--- a/src/Token.h
+++ b/src/Token.h
@@ -86,10 +86,10 @@ public:
 
     std::vector<std::string>::iterator it;
     for (it = NA.begin(); it != NA.end(); ++it) {
-      if ((size_t) (end_ - begin_) != (*it).size())
+      if ((size_t) (end_ - begin_) != it->size())
         continue;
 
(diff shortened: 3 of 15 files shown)


R/tokenizer.R
```diff
@@ -34,7 +34,8 @@ NULL
 
 #' @export
 #' @rdname Tokenizers
-#' @param na String to use for missing values.
+#' @param na Character vector of strings to use for missing values. Set this
```
hadley: I think it would be slightly simpler to use `character()` (i.e. a zero length character vector)` to indicate no missings

And maybe the default value for na should become `c("", "NA")`?

hadley:
This looks great! Apart from the changes mentioned in line, could you please also add a bullet point to `NEWS`, and look into why travis is failing?

R/tokenizer.R
andres-s: Happy to make both changes. `utils::read.csv` has `na.strings = "NA"` by default:
```r
example = "a,b
1,foo
,
"
f = textConnection(example)
read.csv(f, colClasses=c(a="integer", b="character"))
#>    a   b
#> 1  1 foo
#> 2 NA    
close(f)
```
As you can see though, it turns empty values into `NA` where the column type is logical, integer or numeric, so I think extending the same behaviour to character columns by default is pretty reasonable.

andres-s:
Travis *was* failing because the [package version was insufficient](https://travis-ci.org/hadley/readr/builds/69886769#L1752). I cheekily amended the last commit to also bump the version to 0.1.1.9000 and the build is now passing... shall I drop the version back down to 0.1.0.9000 or keep it up? Note that at least one other [open pull request](https://github.com/hadley/readr/pull/183) is failing to build because of the same issue: https://travis-ci.org/hadley/readr/builds/67054495#L1666.

hadley:
Ah, that was because of the CRAN brouhaha. I've bumped in the master branch.

R/tokenizer.R
hadley: Yeah, that makes sense to me too.

andres-s:
OK, I've updated `NEWS`, the default `na` option, changed `NULL` to `character()` and rebased. Let me know if there are any other adjustments to be made and I'll polish them off tomorrow.

NEWS.md
```diff
@@ -17,3 +17,5 @@
 * When guessing field types, and there's no information to go on, use
   character instead of logical (#124, #128).
 
+* Multiple strings can now be specified to indicate missing values (#125).
```
hadley: Maybe also mention the default has changed?

src/Token.h
```diff
@@ -82,14 +82,19 @@ class Token {
...
+  Token& flagNA(std::vector<std::string> NA) {
 
-    if (strncmp(begin_, &NA[0], NA.size()) != 0)
-      return *this;
+    std::vector<std::string>::iterator it;
+    for (it = NA.begin(); it != NA.end(); ++it) {
+      if ((size_t) (end_ - begin_) != (*it).size())
+        continue;
+
+      if (strncmp(begin_, &(*it)[0], (*it).size()) == 0) {
```
hadley: I think `it->size()` would be a bit simpler, and I assume you can simplify the other expression similarly. (`it->data()` maybe?)

src/Tokenizer.cpp
```diff
@@ -14,7 +14,14 @@ TokenizerPtr Tokenizer::create(List spec) {
   if (subclass == "tokenizer_delim") {
     char delim = as<char>(spec["delim"]);
     char quote = as<char>(spec["quote"]);
-    std::string na = as<std::string>(spec["na"]);
+
+    std::vector<std::string> na;
+    if (Rf_isNull(spec["na"])) {
```
hadley: I think you can drop this check now

src/Token.h
andres-s: Huh, I thought I would get a type error because `it->data()` would return a `std::string` pointer, but `strncmp` would expect a `char` pointer, but it works. I guess `std::string` is secretly `char[]` or something.

andres-s:
Thanks for picking up those points, I've made the changes and the code is much cleaner now.

PR #218, opened 2015-07-20
Fix daylight savings bug for unambiguous times. Closes #120

### LINKED ISSUE #120: DateTime fails
x = read_csv("DateTime\n2013-02-10 12:34:56")
fails. I have to supply col_types:
x = read_csv("DateTime\n2013-02-10 12:34:56", col_types="T")

x = read_csv("DateTime\n2013-02-10 12:34:56 JST")
doesn't fail, but the result is a character string, not POSIXct.

andres-s:
Note that this patch does *not* provide unambiguous nor even deterministic parsing of local times for the two hours that overlap at the end of daylight savings, as a consequence of the return value of `mktime()` being undefined for those two hours.

Perhaps the user should be warned/the return value should be fixed when parsing those two problem hours, but it's not clear how best to implement such behaviour.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -40,3 +40,6 @@
 
 * Compound formats %D, %F, %R, %X, %T, %x are now parsed correctly, instead of
   using the ISO8601 parser (#178, @kmillar)
+  
+* Local (non-UTC) times with and without daylight savings are now parsed
+  correctly (#120, @andres-s).
--- a/src/DateTime.h
+++ b/src/DateTime.h
@@ -261,6 +261,10 @@ private:
     tm.tm_hour = hour_;
     tm.tm_min = min_;
     tm.tm_sec = sec_;
+    // The Daylight Saving Time flag (tm_isdst) is greater than zero if Daylight
+    // Saving Time is in effect, zero if Daylight Saving Time is not in effect,
+    // and less than zero if the information is not available.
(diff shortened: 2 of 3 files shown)


hadley:
Oh nice, it's here. Thanks!  Would also mind adding a bullet point to NEWS?

hadley:
It might also be useful to add a comment like "The Daylight Saving Time flag (tm_isdst) is greater than zero if Daylight Saving Time is in effect, zero if Daylight Saving Time is not in effect, and less than zero if the information is not available." (from http://www.cplusplus.com/reference/ctime/tm/)

andres-s:
Amended the commit to add a bullet point to news and the comment to DateTime.h. Let me know if there are any other fixes you would like me to make.

NEWS.md
```diff
@@ -40,3 +40,6 @@
 
 * Compound formats %D, %F, %R, %X, %T, %x are now parsed correctly, instead of
   using the ISO8601 parser (#178, @kmillar)
+  
+* Local (non-UTC) times with and without daylight savings are now parsed
+  correctly (#120).
```
hadley: Can you add you github username here please? (in the same format as above)

andres-s: Ack should've noticed that. Cheers

hadley:
Thanks - much appreciated!

PR #222, opened 2015-07-23
Fixed bug in fwf_empty

hoehleatsu:
There was an unintentional double negotation when checking on the number of columns.
(this is my first pull request, so I don't know if I follow the right procedure here by doing it like this)


hadley:
Thanks! Could you please also add a bullet point to NEWS describing the change?

hoehleatsu:
With pleasure. I added a bullett to NEWS.md.

hadley:
Thanks!

PR #243, opened 2015-09-06
add n_max param to count_fields(). expose skip param to tokenize(), read_lines(), read_file().

npjc:
This PR adds:
- `n_max` argument to `count_fields()` (as implemented for `tokenize()`). 
- Also, add a couple simple tests for parsing fields with `count_fields()`.
- Exposes `skip` argument in `tokenize()` and `read_lines()` and `read_file()`.

Apart from the fact that I personally find these changes useful, I think it would be awesome if the entire interface had `n_max` and `skip` (i.e. would be as consistent as possible). Practically speaking on your end it could mean the documentation side of things everything could `@inheritParams datasource` making it easier to maintain and evolve (i think?).

On that note, my quick look through as to what is missing:

- `read_file()` has no `n_max` argument. 
- `read_rds()` has neither `n_max` nor `skip` arugments.

### DIFF
--- a/R/input.R
+++ b/R/input.R
@@ -4,7 +4,7 @@
 #' @export
 #' @examples
 #' read_file(file.path(R.home(), "COPYING"))
-read_file <- function(file) {
-  ds <- datasource(file)
+read_file <- function(file, skip = 0) {
+  ds <- datasource(file, skip = skip)
   read_file_(ds)
 }
--- a/R/input.R
+++ b/R/input.R
@@ -4,7 +4,7 @@
 #' @export
 #' @examples
 #' read_file(file.path(R.home(), "COPYING"))
-read_file <- function(file, skip = 0) {
-  ds <- datasource(file, skip = skip)
+read_file <- function(file) {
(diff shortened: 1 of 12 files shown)


R/input.R
```diff
@@ -4,7 +4,7 @@
 #' @export
 #' @examples
 #' read_file(file.path(R.home(), "COPYING"))
-read_file <- function(file) {
-  ds <- datasource(file)
+read_file <- function(file, skip = 0) {
```
hadley: Exposing `skip` here doesn't make a lot of sense to me, since `read_file()` is not line based

hadley:
Seems reasonable to me, apart from the one suggestion. I think every function that has some notion of lines should have `n_max` and `skip` params

npjc:
@hadley Good point for `read_file()` have removed as per suggestion (and the use-case for reading _part_ of a file into a single string is probably rare enough that it would't be used anyway). should be good to go.

hadley:
Thanks!

PR #248, opened 2015-09-11
Remove extra datasource in read_delimited()

npjc:
this PR removes a superfluous call to datasource() in read_delimited()..that's it!


npjc:
oh I misunderstood something i see now nevermind closing.

Issue #163, opened 2015-05-04
read compressed files from a URL


katwre:
Here is an example with an error:
```{r}
l <- "http://hgdownload.cse.ucsc.edu/goldenPath/hg19/encodeDCC/wgEncodeHaibMethylRrbs/wgEncodeHaibMethylRrbsImr90UwSitesRep1.bed.gz"
d <- read_delim(l, delim="\t")
Error in collectorsGuess(source, tokenizer, n = 100) : 
  embedded nul in string: '\x89\032\xc7\003\f\xc4#}\xe8\xda\xf5\xbb\x80\xf1\0\023ēo\026O\xe4#\x98Ч\x92\b\xe3\021L\xe4\xf3\xb0\x88\xe3A\006|\xce7澯\xb6n\xeaT\xea\xce\xe7]\\\030\xea|\xb2>$7\xb4\xeb9\xb1ʺ\xf0\xf9u\xa1\031\xd7,~\xbe\177&\xbc\xca\xc5\xea\016\xbd\004t\xbe\x82\xef\xbfR\x8fY_Y\xfa\xfd\xaa\\Ot\x8dH\xc0\x97\xfd\xbb\xbe\027l\xea\xbb\177\xd5Ýw\xa6\v\xe2|K\xbf\xe4\xe7\xd8<\xf9\xad\xeb\x93;Uu\xcbh\003\xf9\xd94\022\x9f\x8b\x95ɦ\037\027\xf2\xbd\xde;\xa7\xa7\xb0\xd2qm3C\xf1\xe7\003\xbf\037#\xed}X\xbd\xb1\x97\xa0b\xaf\xc4J?V\xe8\x85T\xe8U\xf8/\026]\x93I\xeb6\032\x8a?'Iy\xb5\x97U\x8a\xbd\004\025{eV\xe5\017/\xa4B\xaf\xce\177Q\x97Ð]\002\xf2j\001;\x89\x94\xae\xb1AT\x82\xe2\xd7\xe7#\032\x9a\xe5\xe7e\025\x86%\xa1 \xac\x91D\xca|\x84\177]\xc8\xfb\x80Τ\xc2\xca\177\025'
```

maehler:
I just experienced the same issue:

    > read_tsv("ftp://plantgenie.org/Data/PopGenIE/Populus_tremula/v1.0/Annotation/Potra01_GO.tsv.gz")
    Error in collectorsGuess(source, tokenizer, n = 100) : 
[...]

PR #251, opened 2015-09-12
Test updates

DanRuderman:
I've added tests for `read_file` (and `datasource` implicitly). These cover local files, https, raw, etc. Testing fails on reading a compressed file via a URL, as in issue #163. To fix, `datasource()` would need to detect that the string is both a URL and has a compressed file extension; then it would have to wrap the `curl` with the appropriate reader.


DanRuderman:
Hard-coded `https://` in `test-read-file.R` will need to be adjusted.

hadley:
Thanks!

DanRuderman:
Sure. Feel free to assign more...

PR #259, opened 2015-09-14
Update locales.Rmd

ijlyttle:
I could not get this code to work using `"tz"`, but I could using `"tzone"` - it seems half of R uses one convention and the other half the other... grrr..

I don't know whether you meant "instance" or "instants", so I made a presumptuous change.


hadley:
Thanks!

Issue #250, opened 2015-09-11
set column types by regex?


kcandrews:
My data has hundreds of columns, many of them having the same form, such that a regular expression might be used to select all the columns whose header matches with a particular form and assign them the same column type. I couldn't find any examples of this case. Is this possible now, or would you be willing to add this feature?

dewittpe:
I was wondering the same and I created a function `select_cols` for doing this.  A working example is in [this gist](https://gist.github.com/dewittpe/bcc78fd45b3d4887deb5).

I'd be happy to see this type of function as part of the `readr` package too.  Time to fork and work on the feature.

dewittpe:
While developing this feature, have found two warning messages occurring.  A parsing warning that fewer columns have been read in than in the data source, and that the `col_euro_double()` has been deprecated.  The latter warning is of concern because, in the example below, no explicit call to `col_euro_double()` is made.  (I think it is coming from the call to `col_euro_double()` from within `col_types_standardise()`.

    library(devtools) 
    load_all("~/readr")  # the select-col-by-regex branch of dewittpe/reader

    # Example data 
    input_data <- 
    '"","sid","sex","age","group","proc1","proc2","proc3","proc4","proc5","proc6","proc7","proc8","proc9","proc10","proc11","proc12","procdos1","procdos2","procdos3","procdos4","procdos5","procdos6","procdos7","procdos8","procdos9","procdos10","procdos11","procdos12"
    "1",1,"F",8,"C",763.83,441.6,400.08,708.04,138.69,623.44,700.68,663.61,648.04,421.05,913.26,233.6,4,0,5,14,2,33,7,1,36,28,2,3
    "2",2,"F",13,"A",829.94,492.19,412.07,984.53,226.43,242.11,100.22,295.44,853.11,469.57,643.12,172.24,9,18,5,14,10,14,8,0,10,7,12,9
    "3",3,"M",17,"C",449.29,133.69,458.63,783.58,294.74,423.12,287.71,294.91,776.36,616.12,668.35,517.66,5,7,2,22,30,35,0,7,39,2,15,6
    "4",4,"F",5,"C",716.65,976.18,806.22,609.83,531.45,681.06,939.72,450.05,507.45,630.7,943.64,801.42,12,18,7,28,11,0,38,10,9,16,8,8
[...]

PR #261, opened 2015-09-15
Select cols by regex

dewittpe:
This pull request offers a solution to hadley/readr#250.  The added code allows the user to read in a subset of columns using `readr::read_*` and a new function `col_regex_type()`.  Extended documentation and examples have been included.

This feature is helpful for those with wide data sets and need only to read in a subset of the columns.  Could be preferable to reading in all the data, possibly with less than desirable modes, and then using `dplyr::select() %>% dplyr::mutate()` to par down and convert the data modes.

The suggested feature's coding may need improvement.  By creating a `list` object with an expected attribute, this feature was easy to add to the code base.  However, there may be a better solution that would fit into the overall design of `readr`.  I'm happy to redesign and improve the feature.


hadley:
I'd like to think about this a bit more before settling on a concrete implementation. It seems like it might be nice to be able to use the helpers from `select()` like `starts_with()`, `ends_with()` etc.

PR #280, opened 2015-10-07
Augments documentation

bearloga:
Documentation now informs the user that "**D**"ate is a valid column type when using `col_types` in `read_delim()`, etc.

### DIFF
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -36,7 +36,7 @@ NULL
 #'
 #'   Alternatively, you can use a compact string representation where each
 #'   character represents one column: c = character, d = double, i = integer,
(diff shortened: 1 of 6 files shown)


hadley:
Want to add date time and time too?

bearloga:
Now also includes keys for European-style doubles, numbers, date times, and times.

hadley:
Euro doubles don't exist anymore. Would you mind removing?

bearloga:
Sure, but what about https://github.com/hadley/readr/blob/master/R/col_types.R#L102? Is that going away?

hadley:
That generates a deprecation message

bearloga:
Ah, okay. Done :)

hadley:
Thanks!

PR #312, opened 2015-11-13
Add trim_ws argument to read_delim

noamross:


rudazhan:
I'd like to see this commit merged! `read_csv`, `read_csv2`, and `read_tsv` all have this argument and are supposed to be special cases of the general `read_delim` function. So there's not reason for `read_delim` not to have it.

hadley:
Looks good - can you please add a bullet to news?

noamross:
Added and squashed

hadley:
Thanks!

PR #353, opened 2016-01-19
Fixing of ragged fixed width format files and reading a subset of columns

ghaarsma:
This pull request should fix items 300 and 326. The current fwf format is kind of broken if you want to read a subset of columns.

Ragged fwf (where the last column width if variable) do exists, but they should be a minority. This implementation assumes ragged fwf only when the last end position is NA, Inf or omitted.

```R
x <- '12345A\n67890BBBBBBBBB\n54321C'

col_names <- c('A','B','C')
start <- c(1,3,6)
end   <- c(2,5,6)

# Read all columns, non Ragged
col_positions <- fwf_positions(start,end,col_names)
df1 <- read_fwf(x,col_positions = col_positions);df1

# Read subset of columns, it works!
col_positions <- fwf_positions(start[1:2],end[1:2],col_names[1:2])
df2 <- read_fwf(x,col_positions = col_positions);df2

# Read Ragged
col_positions <- fwf_positions(start,end[1:2],col_names)
df3 <- read_fwf(x,col_positions = col_positions);df3

# Read Ragged, alternate way
col_positions <- fwf_positions(start,end=c(2,5,Inf),col_names)
df4 <- read_fwf(x,col_positions = col_positions);df4

# Read Ragged alternate way with fwf_widths
col_positions <- fwf_widths(widths = c(2,3,NA),col_names)
df5 <- read_fwf(x,col_positions = col_positions);df5
```

### DIFF
--- a/R/read_fwf.R
+++ b/R/read_fwf.R
@@ -77,7 +77,13 @@ fwf_widths <- function(widths, col_names = NULL) {
 #' @export
 #' @param start,end Starting and ending (inclusive) positions of each field.
 fwf_positions <- function(start, end, col_names = NULL) {
-  stopifnot(length(start) == length(end))
+
+  # Support for Ragged fwf files. Remove last end element when it is Inf or NA
+  if (is.infinite(end[length(end)]) || is.na(end[length(end)])) {
+    end <- end[-length(end)]
+  }
+
+  stopifnot((length(start)-length(end)) %in% 0:1)
 
   if (is.null(col_names)) {
     col_names <- paste0("X", seq_along(start))
--- a/R/read_fwf.R
+++ b/R/read_fwf.R
@@ -13,8 +13,8 @@
 #' @inheritParams read_delim
 #' @param col_positions Column positions, as created by \code{fwf_empty},
 #'   \code{fwf_widths} or \code{fwf_positions}. To read in only selected fields,
-#'   use \code{fwf_positions}. The width of the last column will be silently
-#'   extended to the next line break.
+#'   use \code{fwf_positions}. If the width of the last column is variable (a
+#'   ragged fwf file), supply the last end position as NA, Inf or simply ommit it.
 #' @export
 #' @examples
 #' fwf_sample <- system.file("extdata/fwf-sample.txt", package = "readr")
@@ -65,7 +65,8 @@ fwf_empty <- function(file, skip = 0, col_names = NULL) {
 
 #' @rdname read_fwf
 #' @export
-#' @param widths Width of each field.
+#' @param widths Width of each field. Use NA or Inf as width of last field when
+#'    reading a ragged fwf file.
 #' @param col_names Either NULL, or a character vector column names.
 fwf_widths <- function(widths, col_names = NULL) {
   pos <- cumsum(c(1, widths))
@@ -76,6 +77,7 @@ fwf_widths <- function(widths, col_names = NULL) {
 #' @rdname read_fwf
 #' @export
 #' @param start,end Starting and ending (inclusive) positions of each field.
+#'    Use NA or Inf as last end field when reading a ragged fwf file.
 fwf_positions <- function(start, end, col_names = NULL) {
 
   # Support for Ragged fwf files. Remove last end element when it is Inf or NA
--- a/R/read_fwf.R
+++ b/R/read_fwf.R
@@ -77,7 +77,7 @@ fwf_widths <- function(widths, col_names = NULL) {
 #' @rdname read_fwf
 #' @export
 #' @param start,end Starting and ending (inclusive) positions of each field.
-#'    Use NA or Inf as last end field when reading a ragged fwf file.
+#'    Use NA as last end field when reading a ragged fwf file.
 fwf_positions <- function(start, end, col_names = NULL) {
 
   stopifnot(length(start) == length(end))
--- a/R/read_fwf.R
+++ b/R/read_fwf.R
@@ -14,7 +14,7 @@
 #' @param col_positions Column positions, as created by \code{fwf_empty},
 #'   \code{fwf_widths} or \code{fwf_positions}. To read in only selected fields,
 #'   use \code{fwf_positions}. If the width of the last column is variable (a
-#'   ragged fwf file), supply the last end position as NA, Inf or simply ommit it.
+#'   ragged fwf file), supply the last end position as NA.
 #' @export
 #' @examples
 #' fwf_sample <- system.file("extdata/fwf-sample.txt", package = "readr")
@@ -65,7 +65,7 @@ fwf_empty <- function(file, skip = 0, col_names = NULL) {
 
 #' @rdname read_fwf
 #' @export
-#' @param widths Width of each field. Use NA or Inf as width of last field when
+#' @param widths Width of each field. Use NA as width of last field when
 #'    reading a ragged fwf file.
 #' @param col_names Either NULL, or a character vector column names.
 fwf_widths <- function(widths, col_names = NULL) {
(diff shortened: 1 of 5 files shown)


jschoeley:
Good solution! Gives the user the power to *explicitly* specify ragged data via the column position parameter. Serves the stated purpose of `fwf_positions` -- "to read in only selected fields" -- much better than the current implementation.

Also the single failing check is a positive sign as it checks for the problematic `fwf_positions` behavior that @ghaarsma fixes in this pull request.

R/read_fwf.R
```diff
@@ -13,8 +13,8 @@
 #' @inheritParams read_delim
 #' @param col_positions Column positions, as created by \code{fwf_empty},
 #'   \code{fwf_widths} or \code{fwf_positions}. To read in only selected fields,
-#'   use \code{fwf_positions}. The width of the last column will be silently
-#'   extended to the next line break.
+#'   use \code{fwf_positions}. If the width of the last column is variable (a
+#'   ragged fwf file), supply the last end position as NA, Inf or simply ommit it.
```
hadley: I think you only need to use a single sentinel value here, and I'd recommend sticking with `Inf`.

R/read_fwf.R
```diff
@@ -76,8 +77,15 @@ fwf_widths <- function(widths, col_names = NULL) {
 #' @rdname read_fwf
 #' @export
 #' @param start,end Starting and ending (inclusive) positions of each field.
+#'    Use NA or Inf as last end field when reading a ragged fwf file.
 fwf_positions <- function(start, end, col_names = NULL) {
-  stopifnot(length(start) == length(end))
```
hadley: I think you should leave this check

R/read_fwf.R
hadley: Actually NA would be easier since because `Inf` is only available in doubles, not integers

src/TokenizerFwf.cpp
```diff
@@ -79,12 +79,16 @@ TokenizerFwf::TokenizerFwf(const std::vector<int>& beginOffset, const std::vecto
   cols_(beginOffset.size()),
   moreTokens_(false)
 {
-  if (beginOffset_.size() != endOffset_.size())
+  // File is assumed to be ragged (last column can have variable width)
+  // when length of end vector is one shorter then begin vector.
+  isRagged_ = beginOffset_.size()==(endOffset_.size()+1L);
```
hadley: This will get simpler once you are checking for explicit NA

src/TokenizerFwf.cpp
```diff
@@ -133,27 +137,30 @@ Token TokenizerFwf::nextToken() {
...
-  // Find the end of the field, stopping for newlines
-  for(int i = 0; i < width; ++i) {
-    if (fieldEnd == end_ || *fieldEnd == '\n' || *fieldEnd == '\r') {
-      warn(row_, col_, tfm::format("%i chars", width), tfm::format("%i", i));
-
-      tooShort = true;
-      break;
+  if (lastCol && isRagged_) {
+    // Last column is ragged, so read until end of line (ignoring width)
+    if (lastCol) {
```
hadley: Won't this always be true?

hadley:
I think this is a reasonable approach, although it still needs quite a bit of work.  @ghaarsma are you interested in continuing to work on it?

@jschoeley A failing test is not a good sign for a PR - tests need to be updated too.

src/TokenizerFwf.cpp
```diff
@@ -164,6 +171,14 @@ Token TokenizerFwf::nextToken() {
     row_++;
     col_ = 0;
 
+    // Proceed to the end of the line. This is needed in case the last column
+    // in the file is not being read.
+    while(fieldEnd != end_ && *fieldEnd != '\r' && *fieldEnd != '\n') {
```
hadley: Are you sure this won't end up accidentally skipping blank lines?

src/TokenizerFwf.cpp
ghaarsma: I agree that we should only use one single sentinel value (NA).
I think the code will be indeed cleaner if we check for explicit NA inside the cpp code and leave the original length checks on start and end in the R and cpp code.

src/TokenizerFwf.cpp
ghaarsma: Yes It is. The `if (lastCol) {}` check should be removed.

src/TokenizerFwf.cpp
ghaarsma: Yes there is room for improvement.
1. You don't have to proceed to the end of line if you are short. `tooShort = true` (you are already there).
2. You don't have to proceed to the end of line if format is ragged (reading the ragged column will do this).
3. However if not short and not ragged then you will not be at the EOL if the user does not wish to read the final column. In that case you have to proceed to EOL and you don't know the width, therefore the while loop.

ghaarsma:
Hadley,
I made a few modifications based on your comments and did a push to https://github.com/ghaarsma/readr. Not sure about the exact work process, so let me know if you need anything else. Code is a little cleaner now. Thanks for the feedback.

The (only) way to indicate a Ragged fixed width format is to have the last position of the end vector as NA. I have checked it with empty lines inside the file (it pushes in a row of NA's), which I assume is the intended behavior.

Some simple tests
```r
x <- '12345A\n67890BBBBBBBBB\n54321C'

col_names <- c('A','B','C')
start <- c(1,3,6)
end   <- c(2,5,6)

# Read all columns, non Ragged
col_positions <- fwf_positions(start,end,col_names)
df1 <- read_fwf(x,col_positions = col_positions);df1

# Read subset of columns, it works!
col_positions <- fwf_positions(start[1:2],end[1:2],col_names[1:2])
df2 <- read_fwf(x,col_positions = col_positions);df2

# Read Ragged
col_positions <- fwf_positions(start,end=c(2,5,NA),col_names)
df4 <- read_fwf(x,col_positions = col_positions);df4

# Read Ragged alternate way with fwf_widths
col_positions <- fwf_widths(widths = c(2,3,NA),col_names)
df5 <- read_fwf(x,col_positions = col_positions);df5
```

R/read_fwf.R
```diff
@@ -76,7 +77,9 @@ fwf_widths <- function(widths, col_names = NULL) {
 #' @rdname read_fwf
 #' @export
 #' @param start,end Starting and ending (inclusive) positions of each field.
+#'    Use NA or Inf as last end field when reading a ragged fwf file.
```
hadley: I think you missed the `Inf` here

src/TokenizerFwf.cpp
```diff
@@ -81,10 +81,14 @@ TokenizerFwf::TokenizerFwf(const std::vector<int>& beginOffset, const std::vecto
 {
   if (beginOffset_.size() != endOffset_.size())
     Rcpp::stop("Begin (%i) and end (%i) specifications must have equal length",
-      beginOffset_.size(), endOffset_.size());
+               beginOffset_.size(), endOffset_.size());
+
+  // File is assumed to be ragged (last column can have variable width)
+  // when the last element of endOffset_ is NA
+  isRagged_ = endOffset_[endOffset_.size()-1L] == NA_INTEGER;
```
hadley: Can you put spaces around `-` please?

src/TokenizerFwf.h
```diff
@@ -12,7 +12,7 @@ class TokenizerFwf : public Tokenizer {
 
   SourceIterator begin_, curLine_, end_;
   int row_, col_, cols_, max_;
-  bool moreTokens_;
+  bool moreTokens_,isRagged_;
```
hadley: Missing space after `,`

hadley:
This is looking good.  Can you please add those examples as unit tests, and add a bullet point to NEWS? (it should include the original issue number and your github user name)

src/TokenizerFwf.cpp
ghaarsma: Fixed

src/TokenizerFwf.h
ghaarsma: Fixed

ghaarsma:
Added new unit tests and a bullet point to the NEWS.md. Let me know if this is all correct. Some of the work process is a first time for me.

hadley:
This looks great!

There's one last thing for you to attempt - can you please rebase/merge to bring your branch up to date with the other changes? If you get stuck, don't worry too much, as I can do it by hand, but it's a good learning experience for future PRs if you want to give it a shot.

ghaarsma:
Hadley,

I tried to follow: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request. Had some minor issues, but I think I got it. Please check when merging the pull request.

hadley:
Perfect - thanks!

PR #360, opened 2016-02-04
write_csv: return original data-frame invisibly

tjmahr:
`write_csv` was returning the path, not the data-frame. `write_delim` wasn't returning a data-frame, but a list of values. Fixed both.

I also added unit tests. It writes to a dummy csv in the testhat folder. I wasn't sure it this is the best way to test file-writing in an R package or if skip_cran() would be required.


tjmahr:
Wrote an issue to accompany this request #363

tests/testthat/test-write-delim.R
```diff
@@ -3,12 +3,21 @@ context("write_delim")
...
+  csv <- write_delim(data.frame(x), "", delim = ",", col_names = FALSE)
   expect_equal(csv, 'a\n\",\"\n')
-  ssv <- write_delim(data.frame(x), "", delim = " ",col_names = FALSE)
+  ssv <- write_delim(data.frame(x), "", delim = " ", col_names = FALSE)
   expect_equal(ssv, 'a\n,\n')
 })
 
+test_that("input is invisibly returned", {
+  csv_in <- read_csv("basic-df.csv")
+  csv_out <- write_csv(csv_in, "basic-df-write.csv")
```
hadley: Can you please just write to `tempfile()`?

NEWS.md
```diff
@@ -170,7 +170,7 @@ As well as improvements to the parser, I've also made a number of tweaks to the
 * `type_convert()` gains `NA` and `trim_ws` arguments, and removes missing
   values before determining column types.
 
-* `write_csv()`, `write_delim()`, and `write_rds()` all invisably return their
```
hadley: This should be a new bullet, not a fix to a previous release

hadley:
Actually, I think this needs a bit more tweaking, so it'll be easier for me to just do it myself. Thanks for brining it to my attention!

PR #366, opened 2016-02-18
Require data.frame input for type_convert()

jennybc:
Otherwise, when input is character, the function makes it impressively far before failing with a cryptic message:

``` r
library(readr)
not_df <- matrix(letters[1:12], nrow = 4,
                 dimnames = list(NULL, paste0("V", 1:3)))
type_convert(not_df)
#> Error: `col_names` must be TRUE, FALSE or a character vector
type_convert(not_df, col_names = TRUE)
#> Error in type_convert(not_df, col_names = TRUE): unused argument (col_names = TRUE)
```
Of course it would also be pretty sweet if `type_convert()` accepted a character matrix as input.

### DIFF
--- a/R/type_convert.R
+++ b/R/type_convert.R
@@ -21,6 +21,7 @@
 #' str(type_convert(df))
 type_convert <- function(df, col_types = NULL, na = c("", "NA"), trim_ws = TRUE,
                          locale = default_locale()) {
+  stopifnot(is.data.frame(df))
   is_character <- vapply(df, is.character, logical(1))
 
   char_cols <- df[is_character]
--- a/tests/testthat/test-type-convert.R
+++ b/tests/testthat/test-type-convert.R
@@ -6,3 +6,8 @@ test_that("missing values removed before guessing col type", {
 
   expect_equal(df2$x, c(NA, 10L))
 })
+
+test_that("requires data.frame input", {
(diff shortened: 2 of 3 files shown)


R/type_convert.R
```diff
@@ -21,6 +21,7 @@
 #' str(type_convert(df))
 type_convert <- function(df, col_types = NULL, na = c("", "NA"), trim_ws = TRUE,
                          locale = default_locale()) {
+  stopifnot(inherits(df, "data.frame"))
```
hadley: I think `is.data.frame()` is a better check.

jennybc:
Changed check to `is.data.frame()`.

tests/testthat/test-type-convert.R
```diff
@@ -6,3 +6,8 @@ test_that("missing values removed before guessing col type", {
 
   expect_equal(df2$x, c(NA, 10L))
 })
+
+test_that("requires data.frame input", {
+  not_df <- matrix(letters[1:4], nrow = 2)
+  expect_error(type_convert(not_df))
```
hadley: For future reference, I think it's bad practice to use `expect_error()` without a string because (e.g) `expect_error(type_c0nvert(not_df))`

hadley:
Thanks!

Issue #345, opened 2016-01-05
parsing something like 12:05 PM


johngoldin:
I had some data from a blood pressure cuff with some readings such as "12:05 PM".  parse_time doesn't like that:
> parse_time(c("0:05 PM", "12:05 PM"), format = "%H:%M %p")
Warning: 1 parsing failure.
row col   expected   actual
  2  -- valid date 12:05 PM
[1] 12:05:00 NA  

parse_time has a point. I understand that technically it should be 0:05 PM.  But a bit of googling can demonstrate that there's a fair amount of data out there that doesn't follow the rules.  See http://www.timeanddate.com/time/am-and-pm.html.  One can find lots of examples of schedules that correctly or incorrectly have times such as 12:05 PM. What they mean isn't really ambiguous even if it's technically incorrect. Given that the philosophy behind readr is to smooth out data processing problems, it would be best if parse_time would gracefully handle things like 12:05 PM.

Dulani:
@johngoldin , I just ran into this too: 
`parse_datetime("12/30/2015 11:59:00 PM","%m/%d/%Y %H:%M:%S %p") #works`
`parse_datetime("12/30/2015 12:01:00 PM","%m/%d/%Y %H:%M:%S %p") #error`  

From my perspective, 12:XX PM is a perfectly valid time if you aren't using a 24 hour clock (which is what the PM says to me).  So, I think this is truly a bug, not a result from strictly parsing something that is ambiguous.

Dulani:
Just realized that this is the same issue as #341.

jimhester:
With https://github.com/hadley/readr/pull/426 and changing `%H` to `%I` as expected this now work as expected

``` r
parse_datetime("12/30/2015 12:01:00 PM","%m/%d/%Y %I:%M:%S %p")
#> [1] "2015-12-30 12:01:00 UTC"
```

Issue #341, opened 2015-12-29
datetime variables in the format %H:%M:%OS %p that are between 12.00 PM and 12.59 PM


nevermiss:
Hi Hadley

I’m having trouble parsing datetime variables in the format %H:%M:%OS %p that are between 12.00 PM and 12.59 PM. All other dates work fine. Here is an example:

parse_datetime(“06/JUN/15 06:49:00.000000000 PM”, “%d/%b/%y %H:%M:%OS %p”) # Works
parse_datetime(“06/JUN/15 12:49:00.000000000 PM”, “%d/%b/%y %H:%M:%OS %p”) # Errors
parse_datetime(“06/JUN/15 12:49:00.000000000”, “%d/%b/%y %H:%M:%OS”) # Removing the PM Works

I’ve been removing the ‘PM’ as a work-around for now, but would be great to hear your thoughts.

Regardless, this is an amazing package so thank you for it!

jimhester:
This is now fixed by supplying '%I' instead of '%H', which should not be used with '%p'. from `?strptime` entry for '%p'

> ‘%p’ AM/PM indicator in the locale.  Used in conjunction with ‘%I’
          and *not* with ‘%H’.

PR #367, opened 2016-02-19
Fixed bug when parsing 12 hour AM/PM dates.

Dulani:
This fixes/addresses the issues #341 and #345.  This is my first pull request, so please scrutinize it closely.

### DIFF
--- a/src/DateTimeParser.h
+++ b/src/DateTimeParser.h
@@ -260,7 +260,7 @@ public:
 
 private:
   int hour() {
-    return hour_ + (amPm_ == 1 ? 12 : 0);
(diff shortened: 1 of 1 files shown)


hadley:
Could you please supply some unit tests too?  I need to think a bit whether that logic is correct.

jimhester:
This was fixed as part of https://github.com/hadley/readr/pull/426. In particular `%H` is not supposed to be used with `%p`. From `?strptime`

> ‘%p’ AM/PM indicator in the locale.  Used in conjunction with ‘%I’
          and *not* with ‘%H’.  An empty string in some locales (and
          the behaviour is undefined if used for input in such a
          locale).

Issue #298, opened 2015-10-22
Column detection specification problem in read_fwf


adamdsmith:
Since 0.2.0, `read_fwf` is having trouble parsing some text files that it previously (0.1.1) handled without any problem.

For example:
```
text <- "F Latitude   Longitude   Alt(m)   Date       Time     S\r\nT N37.6 W089.0    201.1 1969-12-31 20:54:04 1\r\nT N37.6 W089.0    201.1 1969-12-31 20:54:05 0\r\n"

col_pos <- fwf_empty(text, skip = 1)
```

As expected, there are seven columns.

```
col_pos         
$begin
[1]  0  2  8 18 24 35 44

$end
[1]  1  7 14 23 34 43 45

$col_names
[1] "X1" "X2" "X3" "X4" "X5" "X6" "X7"
```

`read_fwf` works with readr 0.1.1:
```
works <- read_fwf(text, col_pos, col_types = "_ccncc_", skip = 1)
works
     X2     X3    X4         X5       X6
1 N37.6 W089.0 201.1 1969-12-31 20:54:04
2 N37.6 W089.0 201.1 1969-12-31 20:54:05
```

`read_fwf` has parsing issues in readr 0.2.0 through 0.2.2.  In this case it
tacks on an extra column of `NA`s: 
```
doesnt <- read_fwf(text, col_pos, col_types = "_ccncc_", skip = 1)

Warning: 2 parsing failures.
row col  expected    actual
1  -- 8 columns 7 columns
2  -- 8 columns 7 columns
Warning message:
    Insufficient `col_types`. Guessing 2 columns.

doesnt
     X2     X3    X4         X5       X6   NA
1 N37.6 W089.0 201.1 1969-12-31 20:54:04 <NA>
2 N37.6 W089.0 201.1 1969-12-31 20:54:05 <NA> 
```

adamdsmith:
Sorry, not able to attempt an adequate PR, but the following changes to `col_spec_standardise` in `R/col_types.R` fixed my specific issue documented above.  It also seems to work with every specification I can throw at it.

Specifically, here is the modified code (lines 161-183 of `col_types.R`) , with the two changed lines marked `### CHANGED` above the `### ORIGINAL` line:

    } else if (is.null(type_names) && !guessed_names) {
      # unnamed types & names supplied: match non-skipped columns
      skipped <- vapply(spec$cols, inherits, "collector_skip",
                        FUN.VALUE = logical(1))
      n_read <- sum(!skipped)
[...]

PR #372, opened 2016-02-24
Subset col_types when there are too many; fixes #371

### LINKED ISSUE #371: unexpected result when there are too many col_types
``` r
library(readr)
```

Too few col\_types works as warning says.

``` r
(too_few <- read_csv("v1,v2\n1,2", col_types = "i"))
#> Warning: Unnamed `col_types` should have the same length as `col_names`.
#> Using smaller of the two.
#> Warning: 1 parsing failure.
#> row col  expected    actual
#>   1  -- 1 columns 2 columns
#>   v1
#> 1  1
```

But too many col\_types gives a weird result.

``` r
(too_many <- read_csv("v1,v2\n1,2", col_types = "iii"))
#> Warning: Unnamed `col_types` should have the same length as `col_names`.
#> Using smaller of the two.
#> Warning: 1 parsing failure.
#> row col  expected    actual
#>   1  -- 1 columns 2 columns
#>   v1
#> 1  1
problems(too_many)
#>   row  col  expected    actual
#> 1   1 <NA> 1 columns 2 columns
```

In the PR, I can fix the behavior but then I get no `problems()`, which doesn't seem correct either.

jennybc:
The result is now what I expect, but seems like there should be problems? I don't know how to fix that.

``` r
(too_many <- read_csv("v1,v2\n1,2", col_types = "iii"))
#> Warning: Unnamed `col_types` should have the same length as `col_names`.
#> Using smaller of the two.
#>   v1 v2
#> 1  1  2
problems(too_many)
#> [1] row      col      expected actual  
#> <0 rows> (or 0-length row.names)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,8 +13,9 @@
 
 * Negative column widths are now allowed in `fwf_widths()` to facilitate
   compatibility with the `widths` argument in `read.fwf()`. (#380, @leeper)
-* `type_covert()` now accepts only `NULL` or a `cols` specification for
+* `type_convert()` now accepts only `NULL` or a `cols` specification for
   `col_types` (#369, @jimhester).
+* If `col_types` is too long, it is subsetted correctly. (#372, @jennybc)
 * `read_file()`, `read_lines()` and `read_csv()` now return empty objects
   rather than signaling an error when run on an empty file (#356, @jimhester).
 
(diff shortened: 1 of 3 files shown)


hadley:
Is this the same problem as #298?

hadley:
@jimhester can you please review?

jimhester:
LGTM and looks like the proper fix, @jennybc if you could just add a note to NEWS.md I think this can be merged.

#298 seems to be a separate problem, this change does not fix that issue.

jennybc:
Done.

PR #390, opened 2016-04-11
Implemented "Truncated division" instead of  "Floored division"

mbask:
I do not know if it is useful to have `format.time` correctly deal with negative count of seconds (currently it always results in an inflated hh:mm:ss conversion): `format.time(-15300)` returns "-05:45:00" (because `-15300%/%3600L` is rounded to -5) instead of "-04:15:00".
It looks like R %/% arithmetic operator implements "floored division", ie from wikipedia (https://en.wikipedia.org/wiki/Modulo_operation):  

> Donald Knuth[7] described floored division where the quotient is defined by the floor function q = ⌊a/n⌋ and thus according to equation (1) the remainder would have the same sign as the divisor. Due to the floor function, the quotient is always rounded downwards, even if it is already negative.

The pull request implements "truncated division", where the quotient is defined by truncation q = trunc(
a/n).

### DIFF
--- a/R/time.R
+++ b/R/time.R
@@ -1,9 +1,9 @@
 #' @export
 format.time <- function(x, ...) {
   x <- as.integer(x)
-  h <- x %/% 3600L
-  m <- (x - h * 3600L) %/% 60L
-  s <- x %% 60L
+  h <- trunc(x / 3600L)
+  m <- abs(x - h * 3600L) %/% 60L
+  s <- abs(x) %% 60L
 
   hms <- paste0(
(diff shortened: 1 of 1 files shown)


hadley:
This should be fixed as part of the move to the hms package.

@krlmlr are you already familiar with this problem for negative times?

krlmlr:
In hms from CRAN, I see

```
> hms::hms(-15300)
-04:15:00
```

The dev version adds support for fractional seconds and times > 24 h.

PR #406, opened 2016-05-13
Allow for lowercase and 0/1 logical values

jeroenooms:
It is not uncommon that boolean values are noted as `0`/`1` or lowercase `true`/`false` in csv.


hadley:
LGTM - if you add a bullet to news I'll merge.

PR #407, opened 2016-05-13
Support parsing year only dates

jeroenooms:
Support parsing a year-only e.g. `%Y` format:

```r
read_csv("x,y\n2014,2014\n", col_types = list(col_date("%Y"), col_datetime("%Y")))
```

It's a bit of an edge case, but if the user specifically sets the format to `%Y` this should work.

### DIFF
--- a/tests/testthat/test-parsing-datetime.R
+++ b/tests/testthat/test-parsing-datetime.R
@@ -70,7 +70,7 @@ test_that("failed parsing returns NA", {
 })
 
 test_that("invalid specs returns NA", {
-  expect_warning(x <- parse_datetime("2010-02-02", "%Y-%m-%m"))
+  expect_warning(x <- parse_datetime("2010-02-20", "%Y-%m-%m"))
   expect_equal(is.na(x), TRUE)
   expect_equal(n_problems(x), 1)
 })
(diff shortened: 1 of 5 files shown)


hadley:
This will also make `parse_datetime("2010-01", "%Y-%m")`, work, right?

I think it also needs some documentation somewhere explaining that missing month and day will be set to 1, and missing hour, minute, and second will be set to 0.

tests/testthat/test-parsing-datetime.R
```diff
@@ -81,6 +81,10 @@ test_that("ISO8601 partial dates are not parsed", {
   expect_equal(n_problems(parse_datetime("2001-01")), 1)
 })
 
+test_that("Year only gets parsed", {
+  expect_equal(parse_datetime("2010", "%Y"), r_parse("2010-01-01", "%Y-%m-%d"))
```
hadley: Maybe better to compare to `ISODate(2010, 01, 01)`?

jeroenooms:
OK I have updated the PR. Somewhat related: currently `year_` defaults to `-1` which is kind of odd:

```r
parse_date('10-10', '%m-%d')
```

We should probably either default to `1970` or make it required.

hadley:
Yeah, that should be an error. Want to tackle it while you're in there?

jeroenooms:
OK done (in this same pr)

PR #414, opened 2016-06-03
add `.default` to the column vignette.

antoine-lizee:
also a small typo (missing bracket)

### DIFF
--- a/vignettes/column-types.Rmd
+++ b/vignettes/column-types.Rmd
@@ -81,12 +81,22 @@ Use the `col_types` argument to override the default choices. There are two ways
     ))
     ```
     
-    Any omitted columns will be parsed automatically, so the previous call is 
-    equivalent to:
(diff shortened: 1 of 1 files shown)


vignettes/column-types.Rmd
```diff
@@ -81,12 +81,22 @@ Use the `col_types` argument to override the default choices. There are two ways
...
+    )
+    ```
+    
+    You can also set a default type that will be used instead of 
+    relying on the automatic detection for columns you don't specify:
+    
+    ```R
+    read_csv("iris.csv", col_types = cols(
+      Species = col_factor(c("setosa", "versicolor", "virginica")),
+      .default = "d")
```
hadley: I'd rather either use all abbreviations or full names in a single call. Would you mind changing?

hadley:
Thanks!

PR #417, opened 2016-06-07
Only skip the first line if data includes col_names

### LINKED ISSUE #333: Length one character vector always guessed as character
I have trouble attaching files, so I just put the content of the files here since they are simple.

I found a surprising behavior using read_csv function in readr package. The format of date is different when the input file has only one row vs. multiple rows.

Could you please take a look and see if this is a bug or user error?

Thanks a lot!

### file1.csv
```
"COL11","COL21","12/09/15"
```
### file2.csv
```
"COL11","COL21","12/09/15"
"COL12","COL22","12/10/15"
"COL13","COL23","12/11/15"
```

### readr_test.R
```
library("readr")
default.locale <- locale(date_names = "en", date_format = "%m/%d/%y",
                         time_format = "%H:%M:%S", decimal_mark = ".", grouping_mark = NULL,
                         tz = "UTC", encoding = "UTF-8", asciify = FALSE)
file1 <- read_csv('file1.csv', locale = default.locale, col_names = FALSE)
colnames(file1) <- c("COL1","COL2","DATE")
View(file1)
class(file1$DATE)
file2 <- read_csv('file2.csv', locale = default.locale, col_names = FALSE)
colnames(file2) <- c("COL1","COL2","DATE")
View(file2)
class(file2$DATE)
print(class(file1$DATE) == class(file2$DATE))
```

jimhester:
The skip line was being used even if `col_names = FALSE`, skipping only when `TRUE` fixes the issue.

Fixes #333

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # readr 0.2.2.9000
 
+* Fix bug when detecting column types for single row files without headers
+  (#333, @jimhester).
+
 * Fix bug in `read_fwf()`, it will now properly read a subset of columns.
   If the final column is ragged, supply an NA as the final end `fwf_positions`
   or final width `fwf_widths` position (#353,@ghaarsma).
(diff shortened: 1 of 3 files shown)


hadley:
Nice - thanks!

PR #418, opened 2016-06-07
single '-' or '.' are now parsed as characters rather than numeric

### LINKED ISSUE #297: read_delim converts "." to "0"
The problem is that "." is parsed as a number like ".0", and as a result it returns "0".  That's why adding locale = locale(decimal_mark = ",", grouping_mark = "_") to read_delim helps. Here is an example:

```
> file = "chr1\t1\t10\t.\t15\t20\nchr1\t5\t15\t.\t20\t25"
> read_delim(file, delim="\t", col_names=F)
    X1 X2 X3 X4 X5 X6
1 chr1  1 10  0 15 20
2 chr1  5 15  0 20 25
```

Apart of that it converts "+" and "-" to NA or "0" and that I just do not understand, e.g:

```
> file = "chr1\t1\t10\t+\t15\t20\nchr1\t5\t15\t-\t20\t25"
> read_delim(file, delim="\t", col_names=F)
Warning: 1 parsing failure.
row col expected actual
  1  X4 a number       
    X1 X2 X3 X4 X5 X6
1 chr1  1 10 NA 15 20
2 chr1  5 15  0 20 25
```

jimhester:
Explicitly checking for this case seems like the best option. I played around with editing the DFA in parseNumber but it ended up being more hassle than it was worth.

Fixes #297

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Fix bug when detecting column types for single row files without headers
   (#333, @jimhester).
+* Fix bug in `collector_guess()`, single '-' or '.' are now parsed as
+  characters rather than numeric (#297, @jimhester).
 
 * Fix bug in `read_fwf()`, it will now properly read a subset of columns.
   If the final column is ragged, supply an NA as the final end `fwf_positions`
--- a/src/QiParsers.h
+++ b/src/QiParsers.h
@@ -51,6 +51,7 @@ inline bool parseNumber(char decimalMark, char groupingMark, Iterator& first,
 
   double sum = 0, denom = 1;
   NumberState state = STATE_INIT;
+  bool seenNumber = false;
 
   Iterator cur = first;
   for(; cur != last; ++cur) {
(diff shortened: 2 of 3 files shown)


src/CollectorGuess.cpp
```diff
@@ -50,6 +50,10 @@ bool isNumber(const std::string& x, LocaleInfo* pLocale) {
   if (x[0] == '0' && x.size() > 1 && x[1] != pLocale->decimalMark_)
     return false;
 
+  // Only a - or decimal mark
+  if (x.size() == 1 && x[0] == '-' || x[0] == pLocale->decimalMark_)
+    return false;
+
```
hadley: I think the problem with only fixing it here is that then `parse_numeric(c(".". "-"))` might give incorrect values.

hadley:
I think the problem with the DFA is that nothing checks that the DFA actually processes any characters - i.e. we need some way to distinguish a `sum` set to 0 from processing a zero, and `sum` that remains at zero. 

Can you try adding an additional `bool seenNumber` that's set to true in any of the `*cur >= '0' && *cur <= '9'` blocks? Then instead of `return true` you'd `return seenNumber`.

jimhester:
Ok did it that way with https://github.com/hadley/readr/pull/418/commits/73d146c7aecc0a395795d468f6381ab8ede9bf00 and added a test for the `parse_numeric()` case.

hadley:
Thanks!

PR #420, opened 2016-06-07
Parsing negative numbers now returns negative values

### LINKED ISSUE #308: col_number() drops negatives
If I don't request specific column types, then `read_csv` treats negative numbers appropriately.

```r
readr::read_csv("A,B\n-4,5G")

##    A B
## 1 -4 5
```

If I specify that the column is numeric, then `read_csv` drops the negative signs.
```r
readr::read_csv("A,B\n-4,5G", col_types = "nn")

##   A B
## 1 4 5
```

jimhester:
Fixes #308

### DIFF
--- a/src/QiParsers.h
+++ b/src/QiParsers.h
@@ -52,6 +52,7 @@ inline bool parseNumber(char decimalMark, char groupingMark, Iterator& first,
   double sum = 0, denom = 1;
   NumberState state = STATE_INIT;
   bool seenNumber = false;
+  double sign = 1.0;
 
   Iterator cur = first;
   for(; cur != last; ++cur) {
@@ -59,6 +60,7 @@ inline bool parseNumber(char decimalMark, char groupingMark, Iterator& first,
     case STATE_INIT:
       if (*cur == '-') {
         state = STATE_LHS;
+        sign = -1.0;
       } else if (*cur == decimalMark) {
         state = STATE_RHS;
       } else if (*cur >= '0' && *cur <= '9') {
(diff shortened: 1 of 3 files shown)


src/QiParsers.h
```diff
@@ -100,14 +101,14 @@ inline bool parseNumber(char decimalMark, char groupingMark, Iterator& first,
       break;
     case STATE_FIN:
       last = cur++;
-      res = sum;
+      res = negative ? sum * -1 : sum;
```
hadley: Could this just `break` so it returns using the same path as below?

jimhester: It could, I guess the only issue is it wouldn't break out of the for loop, so you would iterate through the rest of the string needlessly (probably doesn't matter).

It could be a goto though...

hadley: Ooooh yeah. Let's leave it as is then.

src/QiParsers.h
```diff
@@ -52,6 +52,7 @@ inline bool parseNumber(char decimalMark, char groupingMark, Iterator& first,
   double sum = 0, denom = 1;
   NumberState state = STATE_INIT;
   bool seenNumber = false;
+  bool negative = *first == '-';
```
hadley: Actually, I think this would be nicer as `int sign = 1`, and then change it in the `*cur == '-'` block below. Then you return `sign * sum`

jimhester: Ok it is so (https://github.com/hadley/readr/pull/420/commits/163f780ce61c1d9379e64dc43f5b77a73fc5b145)

hadley:
LGTM

PR #422, opened 2016-06-07
Dates should be based on doubles rather than integers

### LINKED ISSUE #357: Dates should use double type
This is really esoteric but caused a test of mine to fail in a very puzzling way. Dates created by `readr` are different from those created all other ways (integer vs double).

``` r
library(readr)
xbase <- data.frame(d = as.Date(c("2016-02-02", NA)))
xdplyr <- dplyr::data_frame(d = as.Date(c("2016-02-02", NA)))
xreadr <- read_csv("d\n2016-02-02\nNA")
```

These appear to be the same, up to `tbl_df`-ness.

``` r
str(xbase)
#> 'data.frame':    2 obs. of  1 variable:
#>  $ d: Date, format: "2016-02-02" NA
str(xdplyr)
#> Classes 'tbl_df', 'tbl' and 'data.frame':    2 obs. of  1 variable:
#>  $ d: Date, format: "2016-02-02" NA
str(xreadr)
#> Classes 'tbl_df', 'tbl' and 'data.frame':    2 obs. of  1 variable:
#>  $ d: Date, format: "2016-02-02" NA
```

But the `readr` product stores its dates as integer vs numeric for the other two.

``` r
dput(xbase[2, , drop = FALSE])
#> structure(list(d = structure(NA_real_, class = "Date")), .Names = "d", row.names = 2L, class = "data.frame")
dput(xdplyr[2, ])
#> structure(list(d = structure(NA_real_, class = "Date")), .Names = "d", class = c("tbl_df", 
#> "data.frame"), row.names = c(NA, -1L))
dput(xreadr[2, ])
#> structure(list(d = structure(NA_integer_, class = "Date")), .Names = "d", class = c("tbl_df", 
#> "data.frame"), row.names = c(NA, -1L))
```

And it means even `testthat::expect_equivalent()` fails.

``` r
testthat::expect_equivalent(xdplyr, xreadr)
#> Error: xdplyr not equal to expected
#> Rows in x but not y: 2. Rows in y but not x: 2.
```

jimhester:
Fixes #357

### DIFF
--- a/tests/testthat/test-parsing-datetime.R
+++ b/tests/testthat/test-parsing-datetime.R
@@ -121,6 +121,12 @@ test_that("%Z detects named time zones", {
   )
 })
 
+test_that("parse_date returns a double like as.Date()", {
+  ref <- parse_date("2001-01-01")
+
+  expect_typeof(parse_datetime("2001-01-01"), "double")
+})
+
 # Locales -----------------------------------------------------------------
 
 test_that("locale affects months", {
@@ -212,4 +218,3 @@ test_that("must have either two - or none", {
   expect_equal(collector_guess("200010-10"), "character")
   expect_equal(collector_guess("20001010"), "integer")
 })
-
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Fix bug when guessing type for numbers with a trailing character (#316,
   @jimhester).
+* Fix bug in `parse_date()` constructing dates based on integer vectors rather
+  than numeric vectors (#357, @jimhester).
 
 * Fix bug when detecting column types for single row files without headers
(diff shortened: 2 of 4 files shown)


tests/testthat/test-parsing-datetime.R
```diff
@@ -121,6 +121,12 @@ test_that("%Z detects named time zones", {
   )
 })
 
+test_that("parse_date returns a double like as.Date()", {
+  ref <- parse_date("2001-01-01")
+
+  expect_equal(typeof(parse_datetime("2001-01-01")), typeof(ref))
```
hadley: Slightly clearer to do `expect_typeof(..., "double")`

hadley:
LGTM

PR #424, opened 2016-06-07
collector_guess now guesses time formats as well

### LINKED ISSUE #384: Parsing time with format 04:00:00
Hello. 

When time is entered in the format of HH:MM:SS, I think readr wants to parse this as a date, which in turn leads to a parsing error. 

```
x <- c("04:00:00", "04:30:00", "14:00:22")
	
parse_time(x) # OK

>  [1]  4:00:00  4:30:00 14:00:22

parse_date(x) # Parsing error

>  Warning: 3 parsing failures.
>  row col           expected   actual
>    1  -- date like %Y-%m-%d 04:00:00
>    2  -- date like %Y-%m-%d 04:30:00
>    3  -- date like %Y-%m-%d 14:00:22

write.table(x, "times.csv", sep = ",", row.names = FALSE)

read_csv("times.csv") # Parsing error

>  Warning: 3 parsing failures.
>  row  col   expected   actual
>   1 time valid date 04:00:00
>   2 time valid date 04:30:00
>   3 time valid date 14:00:22

```

The problem is resolved by specifying the column type. 

`read_csv("times.csv", col_types = cols(.default = "c"))`

This means I either need to know what the col_types are in advance (which is often not the case when working with collaborators who have sent me a large number of large spreadsheets), or to make them all character, which then creates more work on the other side and defeats the purpose of the smart things that readr does. type_convert is useful, but will still set the offending column to NA. Is it possible in the presence of this kind of parsing error for readr to set that col_type to character as a default? Perhaps this is working as intended, but I thought I would point it out. 

Many thanks.

jimhester:
One thing to note with this is the default time format is `%H:%M`, so the examples in the issue are still going to be guessed as dates unless they change the `time_format` in the locale (as is done in the tests).

I don't know if we should change the default time format to `%H:%M:%S` or not.

Fixes #384

### DIFF
--- a/src/CollectorGuess.cpp
+++ b/src/CollectorGuess.cpp
@@ -69,6 +69,13 @@ bool isDouble(const std::string& x, LocaleInfo* pLocale) {
   return parseDouble(pLocale->decimalMark_, begin, end, res) && begin == end;
 }
 
+bool isTime(const std::string& x, LocaleInfo* pLocale) {
+  DateTimeParser parser(pLocale);
+
+  parser.setDate(x.c_str());
+  return parser.parseTime();
+}
+
 bool isDate(const std::string& x, LocaleInfo* pLocale) {
   DateTimeParser parser(pLocale);
 
@@ -108,6 +115,8 @@ std::string collectorGuess(CharacterVector input, List locale_) {
     return "double";
   if (canParse(input, isNumber, &locale))
     return "number";
+  if (canParse(input, isTime, &locale))
+    return "time";
   if (canParse(input, isDate, &locale))
     return "date";
   if (canParse(input, isDateTime, &locale))
--- a/src/DateTimeParser.h
+++ b/src/DateTimeParser.h
@@ -81,7 +81,8 @@ public:
   bool parseTime() {
     if (!consumeInteger(2, &hour_))
       return false;
-    consumeThisChar(':');
+    if (!consumeThisChar(':'))
+      return false;
     if (!consumeInteger(2, &min_))
       return false;
     consumeThisChar(':');
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,18 @@
 # readr 0.2.2.9000
 
+* The flexible time parser now requires colons between hours and minutes and
+  optional seconds. (#424, @jimhester)
+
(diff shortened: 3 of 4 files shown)


src/CollectorGuess.cpp
```diff
@@ -69,6 +69,12 @@ bool isDouble(const std::string& x, LocaleInfo* pLocale) {
   return parseDouble(pLocale->decimalMark_, begin, end, res) && begin == end;
 }
 
+bool isTime(const std::string& x, LocaleInfo* pLocale) {
+  DateTimeParser parser(pLocale);
+  parser.setDate(x.c_str());
+  return parser.parse(pLocale->timeFormat_);
```
hadley: Hmmmm, maybe this should default to `parseTime()`, since that's a more flexible default?

jimhester: I think `parseTime()` is too flexible actually. Using it the following tests fail

``` r
collector_guess("0001")
#> [1] "time"
collector_guess("00010203")
#> [1] "time"
```

hadley: Then I'd rather make `parseTime()` less flexible but still correct ;)

jimhester:
I updated `parseTime()` to require `:` in between the digits which fixed the false positives. This is a change in what it accepts though but I think it is necessary for this use case.

src/DateTimeParser.h
```diff
@@ -79,19 +79,18 @@ class DateTimeParser {
...
   // A flexible time parser for the most common formats
   bool parseTime() {
-    if (!consumeInteger(2, &hour_))
-      return false;
-    consumeThisChar(':');
-    if (!consumeInteger(2, &min_))
-      return false;
-    consumeThisChar(':');
-    consumeSeconds(&sec_, NULL);
+    bool valid = consumeInteger(2, &hour_) &&
```
hadley: Do you mind sticking with the previous structure? i.e. `return false` when a condition isn't met - I find that easier to understand.

hadley:
I think you're right that making time stricter is a good idea (but it'll need it's own bullet point in NEWS)

jimhester:
Ok used `if` statement as before and added a news bullet.

hadley:
LGTM

PR #425, opened 2016-06-08
Use hms class rather than custom time class

### LINKED ISSUE #409: Use new hms class
for time values, instead of time.R .

jimhester:
The only real issue is arithmetic on hms objects drops the hms class (https://github.com/rstats-db/hms/issues/18), so I had to coerce back to hms objects in the tests.

Fixes #409


hadley:
LGTM

PR #426, opened 2016-06-08
Support %I for 12 hour times

### LINKED ISSUE #340: col_date or col_datetime does not support %I hour format for AM/PM tagged hours
For example, parsing date "12/31/1991 6:00:00 PM", given date format  = "%m/%d/%Y %I" will return an error. 

readr::parse_datetime("12/31/1991 6:00 PM", "%m/%d/%Y %I")

jimhester:
Fixes #340

### DIFF
--- a/src/DateTimeParser.h
+++ b/src/DateTimeParser.h
@@ -116,7 +116,7 @@ public:
         continue;
       }
 
-      // Any other characters much much exactly.
+      // Any other characters must much exactly.
       if (*formatItr != '%') {
         if (!consumeThisChar(*formatItr))
           return false;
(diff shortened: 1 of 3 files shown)


src/DateTimeParser.h
```diff
@@ -161,6 +161,14 @@ class DateTimeParser {
         if (!consumeInteger(2, &hour_))
           return false;
         break;
+      case 'I': // hour
```
hadley: Do we also want to support single digit times?

jimhester: This seems to support both single and double digit times.

hadley: Ok - can you add a unit test for that case too?

jimhester: Done!

hadley:
LGTM

PR #429, opened 2016-06-09
WIP: parse_deparsed implementation

jimhester:
I still need to do the deparsing when writing to a file, but wanted to get any comments about this end first, it seems to work well. I am not sure if missing values should return `NULL` or `NA`, thoughts?

``` r
deparse2 <- function(x) paste(collapse = "", deparse(x))
parse_deparsed(c(deparse2(head(mtcars)), deparse2(head(iris))))
#> [[1]]
#>                    mpg cyl disp  hp drat    wt  qsec vs am gear carb
#> Mazda RX4         21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> Mazda RX4 Wag     21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> Datsun 710        22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> Hornet 4 Drive    21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
#> Hornet Sportabout 18.7   8  360 175 3.15 3.440 17.02  0  0    3    2
#> Valiant           18.1   6  225 105 2.76 3.460 20.22  1  0    3    1
#> 
#> [[2]]
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 1          5.1         3.5          1.4         0.2  setosa
#> 2          4.9         3.0          1.4         0.2  setosa
#> 3          4.7         3.2          1.3         0.2  setosa
#> 4          4.6         3.1          1.5         0.2  setosa
#> 5          5.0         3.6          1.4         0.2  setosa
#> 6          5.4         3.9          1.7         0.4  setosa
```

### DIFF
--- a/src/Collector.cpp
+++ b/src/Collector.cpp
@@ -39,6 +39,9 @@ CollectorPtr Collector::create(List spec, LocaleInfo* pLocale) {
     bool ordered = as<bool>(spec["ordered"]);
     return CollectorPtr(new CollectorFactor(levels, ordered));
   }
+  if (subclass == "collector_deparsed") {
+    return CollectorPtr(new CollectorDeparsed());
+  }
 
   Rcpp::stop("Unsupported column type");
   return CollectorPtr(new CollectorSkip());
@@ -362,4 +365,37 @@ void CollectorTime::setValue(int i, const Token& t) {
     Rcpp::stop("Invalid token");
   }
 
+}
+
+void CollectorDeparsed::setValue(int i, const Token& t) {
+  switch(t.type()) {
+  case TOKEN_STRING: {
+    boost::container::string buffer;
+    SourceIterators string = t.getString(&buffer);
+    std::string std_string(string.first, string.second);
+    RObject ans = R_NilValue;
+
+    ParseStatus status;
+
+    CharacterVector c(std_string);
+    ExpressionVector r = R_ParseVector(c, -1, &status, R_NilValue);
+    if (status == PARSE_OK) {
+      for(R_len_t i = 0; i < Rf_length(r); i++) {
+        ans = Rf_eval(VECTOR_ELT(r, i), R_GlobalEnv);
+      }
+
+      SET_VECTOR_ELT(column_, i, ans);
+    } else {
+      // parse error
+      Rcpp::stop("Invalid expression %s", c);
+    }
+    return;
+  }
+  case TOKEN_MISSING:
+  case TOKEN_EMPTY:
+    SET_VECTOR_ELT(column_, i, R_NilValue);
+    return;
+  case TOKEN_EOF:
+    Rcpp::stop("Invalid token");
+  }
 }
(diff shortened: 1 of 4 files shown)


src/Collector.cpp
```diff
@@ -363,3 +366,36 @@ void CollectorTime::setValue(int i, const Token& t) {
...
+    ExpressionVector r = R_ParseVector(c, -1, &status, R_NilValue);
+    if (status == PARSE_OK) {
+      for(R_len_t i = 0; i < Rf_length(r); i++) {
+        ans = Rf_eval(VECTOR_ELT(r, i), R_GlobalEnv);
+      }
+
+      SET_VECTOR_ELT(column_, i, ans);
+    } else {
+      // parse error
+      Rcpp::stop("Invalid expression %s", c);
```
hadley: This should probably be a "problem" rather than terminating parsing of the whole file.

src/Collector.cpp
```diff
@@ -363,3 +366,36 @@ void CollectorTime::setValue(int i, const Token& t) {
...
+    std::string std_string(string.first, string.second);
+    RObject ans = R_NilValue;
+
+    ParseStatus status;
+
+    CharacterVector c(std_string);
+    ExpressionVector r = R_ParseVector(c, -1, &status, R_NilValue);
+    if (status == PARSE_OK) {
+      for(R_len_t i = 0; i < Rf_length(r); i++) {
+        ans = Rf_eval(VECTOR_ELT(r, i), R_GlobalEnv);
```
hadley: Hmmmm, I forgot that we'd have to evaluate the string - that makes me a bit concerned. Maybe ASCII serialization would be a better default? (assuming the R C APIs exist)

src/Collector.cpp
jimhester: Yeah agreed I wasn't exactly sure how to do that, but I see there is a `warn()` function to do it.

src/Collector.cpp
jimhester: I don't think evaluating is that bad, you have to opt in for this so presumably you trust the input in that column.

We could just return the unevaluated expressions, but I think it would just make people perform an extra step the vast majority of the time. 

Although `x$col <- lapply(x$col, eval)` is not that bad and would allow people to inspect the expressions before evaluating them.

We could maybe make this an argument as well...

jimhester:
Are you aware of any exported function to do deparsing?

`Rf_deparse1line()` is what would be best, but it doesn't look like it is part of the API :(

hadley:
I was thinking `SEXP R_Unserialize(R_inpstream_t ips)` but you'd probably need to do a bit of digging to figure out what `ips` is supposed to be.  It's probably a connection?  But implementing that is probably overkill.

I just thinking evaluating code from a csv file (even if you've opted in to this) is sufficiently surprising/scary that it basically kills this idea.

jimhester:
Ok I will table it then, we can always re-open in the future if needed.

PR #430, opened 2016-06-09
Return empty objects rather than signaling an error on an empty file.

### LINKED ISSUE #356: read_file on empty file throws error
I'd expect to get an empty string, but instead get:
Error: Cannot read file /tmp/RtmpFO5GFJ/file4dc73de7d6a2.txt

This could be what you designed it to do, but its arguably not what most users would expect from this function, and it should at least be documented (and tested for)

My testcase:

```
f <- tempfile( fileext=".txt" )
cat( "", file=f )
expect_true( file.exists(f) )
expect_that(
    read.file( f ),
    is_identical_to( "" )
)
```

jimhester:
I did the file tests in the R side as I didn't think it was worth pulling in the needed Boost headers to do the equivalent test for an empty file.

Fixes #356

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@
   compatibility with the `widths` argument in `read.fwf()`. (#380, @leeper)
 * `type_covert()` now accepts only `NULL` or a `cols` specification for
   `col_types` (#369, @jimhester).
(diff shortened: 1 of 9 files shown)


hadley:
LGTM.  Might be worth saving empty file in tests directory?

PR #431, opened 2016-06-09
type_covert now accepts only NULL or a cols specification for col_types

### LINKED ISSUE #369: type_convert only touches character columns; doesn't check/subset col_types accordingly
``` r
library(readr)
suppressPackageStartupMessages(library(dplyr))
```

If you can use `read_csv()`, the automatic variable type detection is great.

``` r
read_csv("integer,logical\n1,TRUE\n2,FALSE\n")
#> Source: local data frame [2 x 2]
#> 
#>   integer logical
#>     (int)   (lgl)
#> 1       1    TRUE
#> 2       2   FALSE
```

I understand `type_convert()` is meant to do similar when you're stuck with a *data frame* that needs Hadleyverse type (re-)conversion.

``` r
(df <- data_frame(integer = c(1, 2),
                  logical = c("TRUE", "FALSE")))
#> Source: local data frame [2 x 2]
#> 
#>   integer logical
#>     (dbl)   (chr)
#> 1       1    TRUE
#> 2       2   FALSE
```

Right now, only character columns are re-processed. Could all columns be re-processed? At least when `col_types = NULL`?

``` r
type_convert(df)
#> Source: local data frame [2 x 2]
#> 
#>   integer logical
#>     (dbl)   (lgl)
#> 1       1    TRUE
#> 2       2   FALSE
```

Also, if you specify `col_types`, it's not actually checked/subsetted against the character columns, so you get warnings and an incorrect result.

``` r
type_convert(df, col_types = "il")
#> Warning: Insufficient `col_names`. Adding 1 names.
#> Warning in type_convert_col(char_cols[[i]], col_types[[i]],
#> which(is_character)[i], : [0, 2]: expected an integer, but got 'TRUE'
#> Warning in type_convert_col(char_cols[[i]], col_types[[i]],
#> which(is_character)[i], : [1, 2]: expected an integer, but got 'FALSE'
#> Source: local data frame [2 x 2]
#> 
#>   integer logical
#>     (dbl)   (int)
#> 1       1      NA
#> 2       2      NA
```

Current workaround is to make sure everything is character prior to use of `type_convert()`.

``` r
df[] <- lapply(df, as.character)
type_convert(df)
#> Source: local data frame [2 x 2]
#> 
#>   integer logical
#>     (int)   (lgl)
#> 1       1    TRUE
#> 2       2   FALSE
```

This may be related to what this person was talking about? Hard to tell. <https://github.com/hadley/readr/pull/160>

jimhester:
Another alternative would be to subset the col_types with the is_character logical vector as is already done for the data.frame and col_names. Not sure which way would be more useful to people.

e.g. a diff from this PR

```diff
diff --git a/R/type_convert.R b/R/type_convert.R
index 3a46932..bbd5c6c 100644
--- a/R/type_convert.R
+++ b/R/type_convert.R
@@ -44,8 +44,7 @@ type_convert <- function(df, col_types = NULL, na = c("", "NA"), trim_ws = TRUE,
   })

   if (is.character(col_types)) {
-    stop("`col_types` must be `NULL` or a `cols` specification for `type_convert()`.",
-      call. = FALSE)
+    col_types <- paste(collapse = "", substring(col_types, which(is_character), which(is_character)))
   }

   col_types <- col_spec_standardise(
```

Fixes #369

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@
 
 * Negative column widths are now allowed in `fwf_widths()` to facilitate
   compatibility with the `widths` argument in `read.fwf()`. (#380, @leeper)
+* `type_covert()` now accepts only `NULL` or a `cols` specification for
+  `col_types` (#369, @jimhester).
 
 * time objects returned by `parse_time()` are now `hms` objects rather than a
   custom `time` class (#409, @jimhester).
--- a/R/type_convert.R
+++ b/R/type_convert.R
@@ -5,6 +5,19 @@
 #' then let readr take another stab at parsing it.
 #'
 #' @param df A data frame.
+#' @param col_types One of \code{NULL}, a \code{\link{cols}} specification, or
+#'   a string. See \code{vignette("column-types")} for more details.
+#'
+#'   If \code{NULL}, all column types will be imputed from the first 1000 rows
+#'   on the input. This is convenient (and fast), but not robust. If the
+#'   imputation fails, you'll need to supply the correct types yourself.
+#'
+#'   If a column specification created by \code{\link{cols}}, it must contain
+#'   one "\code{\link{collector}}" for each column. If you only want to read a
+#'   subset of the columns, use \code{\link{cols_only}}.
+#'
+#'   Unlike other functions \code{type_convert} does not allow character
+#'   specificatinos of \code{col_types}.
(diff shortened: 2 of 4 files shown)


hadley:
LGTM

I think specifying character col types is just going to be too confusing.

PR #432, opened 2016-06-10
write_* improvements

### LINKED ISSUE #387: Think about write_csv again
Performance for this is pretty embarrassing:

```R
df <- as.data.frame(matrix(runif(256*65536), nrow = 256))
    
system.time(readr::write_csv(df, "df.csv"))
```

e.g. http://stackoverflow.com/questions/10505605/speeding-up-the-performance-of-write-table/36484339#36484339

jimhester:
The previous implementation as we know was very slow because of the character conversion, on par with `write.csv()`.
```r
set.seed(1)
df <- as.data.frame(matrix(runif(256*2^15), nrow = 256))
system.time(write.csv(df, "/tmp/df1.csv"))
#>    user  system elapsed 
#>  26.074   7.622  33.776
system.time(readr::write_csv(df, "/tmp/df4.csv"))
#>    user  system elapsed 
#>  34.012   0.537  34.589
```

https://github.com/hadley/readr/commit/9c286458d1797b00a0212c3fe374a9841d606c12 just does a little cleanup and turns off converting numeric to character first. This produces valid round trip-able results and is quite a bit faster than converting to character, however all numeric numbers are printed with the maximum amount of precision.

```r
system.time(readr::write_csv(df, "/tmp/df4.csv"))
#>    user  system elapsed 
#>   8.308   0.240   8.588
```

https://github.com/hadley/readr/commit/a67c1d5a7ef415796795e16c9d3903546807c76d uses the grisu3 implementation found at https://github.com/juj/MathGeoLib/blob/master/src/Math/grisu3.c. It is under the [Apache 2 license](http://clb.demon.fi/MathGeoLib/nightly/sourcecode.html#license) so is safe for us to use. This actually gives us quite a bit better performance than the naive approach.

```r
system.time(readr::write_csv(df, "/tmp/df4.csv"))
#>    user  system elapsed 
#>   3.047   0.203   3.265
```

However `data.table:fwrite()` is still faster than any of these methods.
```r
system.time(data.table::fwrite(df, "/tmp/df3.csv"))
#> Your platform/environment has not detected OpenMP support. fwrite() will still work, but slower in single threaded mode.
#>    user  system elapsed 
#>   1.578   0.063   1.651
```

I did some profile sampling  with `R -q -d "valgrind --tool=callgrind --dump-instr=yes --collect-jumps=yes"` and the vast majority of our computational time is doing the formatting, so I am not sure how much more room there is to improve.

Fixes #387

### DIFF
--- a/src/grisu3.c
+++ b/src/grisu3.c
@@ -1,0 +1,361 @@
+/* This file is part of an implementation of the "grisu3" double to string
+	conversion algorithm described in the research paper
+
+	"Printing Floating-Point Numbers Quickly And Accurately with Integers"
+	by Florian Loitsch, available at
+	http://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf */
+
+#include <stdint.h> // uint64_t etc.
+#include <assert.h> // assert
+#include <math.h> // ceil
+#include <stdio.h> // sprintf
+
+#ifdef _MSC_VER
+#pragma warning(disable : 4204) // nonstandard extension used : non-constant aggregate initializer
+#endif
+
+#define D64_SIGN         0x8000000000000000ULL
+#define D64_EXP_MASK     0x7FF0000000000000ULL
+#define D64_FRACT_MASK   0x000FFFFFFFFFFFFFULL
+#define D64_IMPLICIT_ONE 0x0010000000000000ULL
+#define D64_EXP_POS      52
+#define D64_EXP_BIAS     1075
+#define DIYFP_FRACT_SIZE 64
+#define D_1_LOG2_10      0.30102999566398114 // 1 / lg(10)
+#define MIN_TARGET_EXP   -60
+#define MASK32           0xFFFFFFFFULL
+
+#define CAST_U64(d) (*(uint64_t*)&d)
+#define MIN(x,y) ((x) <= (y) ? (x) : (y))
+#define MAX(x,y) ((x) >= (y) ? (x) : (y))
+
+#define MIN_CACHED_EXP -348
+#define CACHED_EXP_STEP 8
+
+typedef struct diy_fp
+{
+	uint64_t f;
+	int e;
+} diy_fp;
+
+typedef struct power
+{
+	uint64_t fract;
+	int16_t b_exp, d_exp;
+} power;
+
+static const power pow_cache[] =
+{
+	{ 0xfa8fd5a0081c0288ULL, -1220, -348 },
+	{ 0xbaaee17fa23ebf76ULL, -1193, -340 },
+	{ 0x8b16fb203055ac76ULL, -1166, -332 },
+	{ 0xcf42894a5dce35eaULL, -1140, -324 },
+	{ 0x9a6bb0aa55653b2dULL, -1113, -316 },
+	{ 0xe61acf033d1a45dfULL, -1087, -308 },
+	{ 0xab70fe17c79ac6caULL, -1060, -300 },
+	{ 0xff77b1fcbebcdc4fULL, -1034, -292 },
+	{ 0xbe5691ef416bd60cULL, -1007, -284 },
+	{ 0x8dd01fad907ffc3cULL,  -980, -276 },
+	{ 0xd3515c2831559a83ULL,  -954, -268 },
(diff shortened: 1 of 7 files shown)


jimhester:
Sample output from grisu3 is identical to current and `write.csv()` on mtcars
```r
readr::write_tsv(head(mtcars), "/dev/stdout")
#> mpg	cyl	disp	hp	drat	wt	qsec	vs	am	gear	carb
#> 21	6	160	110	3.9	2.62	16.46	0	1	4	4
#> 21	6	160	110	3.9	2.875	17.02	0	1	4	4
#> 22.8	4	108	93	3.85	2.32	18.61	1	1	4	1
#> 21.4	6	258	110	3.08	3.215	19.44	1	0	3	1
#> 18.7	8	360	175	3.15	3.44	17.02	0	0	3	2
#> 18.1	6	225	105	2.76	3.46	20.22	1	0	3	1
```

vs setting the precision manually
```r
readr::write_tsv(head(mtcars), "/dev/stdout")
#> mpg	cyl	disp	hp	drat	wt	qsec	vs	am	gear	carb
#> 21	6	160	110	3.8999999999999999	2.6200000000000001	16.460000000000001	0	1	4	4
#> 21	6	160	110	3.8999999999999999	2.875	17.02	0	1	4	4
#> 22.800000000000001	4	108	93	3.8500000000000001	2.3199999999999998	18.609999999999999	1	1	4	1
#> 21.399999999999999	6	258	110	3.0800000000000001	3.2149999999999999	19.440000000000001	1	0	3	1
#> 18.699999999999999	8	360	175	3.1499999999999999	3.4399999999999999	17.02	0	0	3	2
#> 18.100000000000001	6	225	105	2.7599999999999998	3.46	20.219999999999999	1	0	3	1
```

jimhester:
One possible issue is mentioned in the [grisu3 paper](http://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pd), which states the following (emphasis mine)

> With just two extra bits it is difficult to do better than in our
example, but often there exists an integer type with more bits. For
IEEE 754 floating-point numbers, which have a significand size of
53, one can use 64 bit integers, providing 11 extra bits. We have
developed an algorithm Grisu2 that uses these extra bits to shorten
the output. However, even 11 extra bits may not be sufficient in
every case. There are still boundary conditions under which Grisu2
will not be able to produce the shortest representation. Since this
property is often a requirement (see [Steele Jr. and White(2004)]
for some examples) we propose a variant, Grisu3, that detects (and
aborts) when its output may not be the shortest. As a consequence
Grisu3 is incomplete and will fail for some percentage of its input.
**Given 11 extra bits roughly 99.5% are processed correctly and
are thus guaranteed to be optimal (with respect to shortness and
rounding). The remaining 0.5% are rejected and need to be printed
by another printing algorithm (like Dragon4)**.

I need to look a this implementation and see what happens if the input is rejected, but it is reassuring it did not fail with the test inputs, which are random although only between 0-1.

jimhester:
We may also want to incorporate the changes from https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/grisu3_print.h#L207-L228 (also under Apache 2.0) which prefer 'unscientific' notation at the same length and always append a 0 on decimals.

src/grisu3.c
```diff
@@ -0,0 +1,361 @@
+/* This file is part of an implementation of the "grisu3" double to string
```
hadley: Can you include the license too? Might need to copy and paste from somewhere else

hadley:
Also need to update Authors@R

hadley:
You mean C-level formatting or R-level formatting?

But I'm happy with that performance - we don't need to be as fast as `fwrite()`, we just need not to be embarrassingly slow.

jimhester:
C-level formatting is what I meant (after the above changes).

hadley:
Apart from the authorship/license stuff (and news bullet), LGTM. Feel free to merge when you've done those bits.

jimhester:
Added the license and authors to the DESCRIPTION. PTAL briefly to make sure it looks OK and then I can merge this.

hadley:
Looks good - I confirmed that Apache license is compatible with GPL3.

Issue #403, opened 2016-05-11
negative length vectors are not allowed


steverowe:
When reading the .gz version of the data from the Expedia Kaggle competition, I get this error:
Error in .Call("readr_read_connection_", PACKAGE = "readr", con, chunk_size) : 
  negative length vectors are not allowed

File link: https://www.kaggle.com/c/expedia-hotel-recommendations/download/train.csv.gz

hadley:
Can you provide a link that doesn't require a login?

steverowe:
Try https://1drv.ms/u/s!AlqswdwIdiVGttAxS6r3RyBxQlvT0g.  This will let you download the file and then you could open it from there.

I just used a simple
`x <- read_csv("train.csv.gz")`

randomgambit:
Hello! I am getting the same error on a large zipped file. I use the latest version of `dplyr`

```
Error in .Call("readr_read_connection_", PACKAGE = "readr", con, chunk_size) : 
  negative length vectors are not allowed
```

Is there anything I could do to read it with `read_csv` ?

Thanks!

hadley:
@randomgambit please file a reproducible example in a new issue.

randomgambit:
@hadley thanks for the quick feedback. Unfortunately I cannot post the data. This is tick by tick trading data from a major exchange. Is there any test I can do to figure out what is the error? The variable names contain spaces (if that matters) thanks!

PR #433, opened 2016-06-14
Support reading into long vectors

### LINKED ISSUE #309: read_lines craches on big gzipped file
I'm trying to read big text file (~4.5gb in gzipped form, english wikipedia dump)
`read_lines` produce following error:
> Error in read_lines_(ds, locale_ = locale, n_max = n_max, progress = progress) : 
  long vectors not supported yet: ../../src/include/Rinlinedfuns.h:137

EDIT: I'm doing it on machine with 250gb ram, so ram is not an issue.

jimhester:
Fixes #309, #403

I commented out the tests as they are pretty slow to run.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # readr 0.2.2.9000
 
+* Supports reading into long vectors (#309, @jimhester).
+
 * Printing of double values now uses an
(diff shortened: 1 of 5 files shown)


hadley:
LGTM

Can you also check if this fixes #403?

I wish there was some way to test this, but I can't think of anything that isn't going to take ages.

jimhester:
It does fix #403 as well.

PR #436, opened 2016-06-15
Check for BOM when converting

### LINKED ISSUE #263: Support for UTF-8-BOM
Reviewing #208 and #40, it looks like `readr` should support this, but the CRANed 0.1.1 is failing to handle a UTF-8-BOM CSV. If I supply `fileEncoding = UTF-8-BOM` to `read.csv` in base, this file comes in cleanly. Should `readr` be able to handle this?

This file is a MS SQL Server CSV export so...Windows. :wink:

jimhester:
This works for the test case, the main issue is it really only needs to check for this at the beginning of the file (if it is a UTF-8 file). I am still trying to figure out the best place to put this logic, suggestions welcome!

Fixes #263

### DIFF
--- a/src/Source.h
+++ b/src/Source.h
@@ -47,6 +47,57 @@ public:
     return cur;
   }
 
+  static const char* skipBom(const char* begin, const char* end) {
+
+    /* Unicode Byte Order Marks
+       https://en.wikipedia.org/wiki/Byte_order_mark#Representations_of_byte_order_marks_by_encoding
+
+       00 00 FE FF: UTF-32BE
+       FF FE 00 00: UTF-32LE
+       FE FF:       UTF-16BE
+       FF FE:       UTF-16LE
+       EF BB BF:    UTF-8
+   */
+
+    switch(begin[0]) {
+      // UTF-32BE
+      case '\x00':
+        if (end - begin >= 4 && begin[1] == '\x00' && begin[2] == '\xFE' && begin[3] == '\xFF') {
+          return begin + 4;
+        }
+        break;
+
+      // UTF-8
+      case '\xEF':
+        if (end - begin >= 3 && begin[1] == '\xBB' && begin[2] == '\xBF') {
+          return begin + 3;
+        }
+        break;
+
+      // UTF-16BE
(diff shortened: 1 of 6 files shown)


hadley:
Maybe it should be done in the constructor of the sources?

And maybe pull out into `hasUtf8Bom()` function?

jimhester:
ˇI put the logic in a `skipBom()` function which is called from the three Source constructors.

Let me know if you think there is anything else we need to change.

src/Source.h
```diff
@@ -47,6 +47,47 @@ class Source {
     return cur;
   }
 
+  static const char* skipBom(const char* begin) {
+    // Skip any unicode byte order marks
+    switch(begin[0]) {
+      // UTF-32BE
+      case '\x00':
+        if (begin[1] == '\x00' && begin[2] == '\xFE' && begin[3] == '\xFF') {
```
hadley: I think this is a little risky because you might advance past the end of the string

hadley: Otherwise looks good

hadley: Just remembered I faced this problem else where, and the answer is `boost::starts_with`. See e.g. `inComment()`:

```cpp
  static bool inComment(const char* cur, const char* end, const std::string& comment) {
    boost::iterator_range<const char*> haystack(cur, end);
    return boost::starts_with(haystack, comment);
  }
```

hadley: Would also be good to include a comment as to the sources of these byte sequences

jimhester: Using an interator_range will likely have more overhead than doing the comparisons directly, the main benefit of the current code is each byte is compared only once, using any ranged compare would require you to compare the first few bytes more than once for each comparison.

I can make the current safe by passing the end ptr and verifying the length is at least (2 or 4) prior to comparison.

hadley: It's up to you - I don't think this will make much difference to performance either way ;)

jimhester:
You are right we are only calling this once per file. I switched to using `boost::starts_with()` for brevity/readability.

hadley:
LGTM

Issue #314, opened 2015-11-14
Make it easy to catch mutations in the data by emitting col_types string


earino:
One of the nice features of readr is that it introspects column types and gives you data frames with columns of "appropriate" data types. This is cool. It does so heuristically by reading a bunch of rows (but not all) and guessing.

This is very nice for interactive use, however I think it could be made into a neat feature to catch data schema mutation.

I would like to propose a new function which reads a *whole* CSV file, and from the entire file returns a col_types string. A user can then take that string and put it in their production script's read_csv ca;;, so that in the future, if a new data file comes in which has different data types, instead of (for example) silently returning a data frame which has chars where a date once was, it will now give a proper error.

hadley:
This is a great idea :)

jennybc:
If implemented, it might be nice to comply with something like csvy: <http://csvy.org>, <http://dataprotocols.org/json-table-schema/>. The rio package just accommodated csvy (https://github.com/leeper/rio/issues/52).

hadley:
I'm thinking that the default behaviour should be to spit out the code that you can copy and paste into the `col_types` in order to repeat the exact same parsing in the future.

jimhester:
Addressed by #437

drolejoel:
is it OK to add a comment to this completed enhancement, or should I add a new enhancement request?

While spitting out the column specification (I like this) is a great feature as a default, it gets a tad tedious to see the same red text on the screen once the spec has been figured out. My suggestion is to add a "silent = TRUE" option to suppress the column specifications if they are not wanted.
Thank you for your consideration.

hadley:
It's always better to submit a new issue.  (But this one is already handled: just do `col_types = col()`)

drolejoel:
ok  thank you

Issue #304, opened 2015-11-02
col_types specifications different between read_excel and read_csv - vector vs. list/string


sfirke:
I use `read_excel` from readxl often and I was surprised that I couldn't use a vector for `col_types` when calling `read_csv` from readr.  The options of a list or a character string both seem more complex than a character vector, which would be consistent with `read_excel` as well as `read.csv` (and I think many users will come to `read_csv` when updating code from `read.csv` so this will be a common stumbling block).

I know readxl is a different package but is this different approach to `col_types` intentional / would you consider letting `read_csv` take a vector for `col_types`?

jjcad:
+1 `read_*` should support option to specifyc `col_types` the same as `colClasses` in `read.table()`.

hadley:
Part of #237

Issue #237, opened 2015-08-29
Think about how to extract colspec package


hadley:
for use in other data import packages

hadley:
@jennybc do I remember that you worked on this?

jennybc:
Well, I think we can say I "asked about / asked for" such a package.  Specifics that have come up in email (some answers from memory).

Q: Are `NA` variable names going to persist?  Is the "no munging of variable names" here to stay"?
A: No, and yes.

Q: Is it intentional that an all-`NA` column is `NA_character_` and not logical?
A: No, it should be logical.
*Sidebar: above applies to `readr`. `readxl` returns `NA_real_` in the same scenario.*

Q: Could `tidyr::spread()` use `readr::type_convert()` instead of base `type.convert()`?
A: That is the plan, but need to keep an eye on dependencies.

Q: Can we have a column conversion package?
A: Perhaps but it's hard.

Consistency around empty columns.

hadley:
I think I got confused with your col range (?) package.

If we did do this, we might want to pull out the `select` helpers from dplyr since it's nice to be able to select sets of variables with them.

PR #437, opened 2016-06-15
Add a return_spec argument to read functions

jimhester:
This tries to address #314 slightly differently than proposed.

It adds a `return_spec` argument to the read_* functions which rather than read the files just returns the `col_spec` object that is generated before reading the file.

This object can then be used in subsequent calls as the value for the `col_types` argument, which will allow you to enforce a specification on other datasets with the same specifications and such.

These objects can be saved retrieved with `saveRDS()/readRDS()` and already have formatting methods written.

We could also write a function to take the object and produce an expression that could be used to recreate the object, but I am not sure that is more useful or informative than just working with the objects directly.


hadley:
Reading this makes me realise that I forgot to describe an important part of the workflow that I was imagining. I was thinking you'd often run it once to make the col-guesses explicit, and then you'd modify that code/text to fix any detection errors and specify the correct columns types.  So I think that makes having a code/text rendering quite important.

hadley:
Also I think to start with I'd rather have this be a separate function, i.e. `spec_csv`, `spec_tsv` etc.

jimhester:
I feel like that workflow works better with this implementation than the proposed one, it is very easy to subset the object and change the columns after returning the col_spec object. I also added a `as.character()` method to print an R expression that would recapitulate the spec object.

``` r
write_csv(mtcars, "mtcars.csv")
spec <- read_csv("mtcars.csv", return_spec = TRUE)
spec
#> <col_spec>
#> * mpg: double
#> * cyl: integer
#> * disp: double
#> * hp: integer
#> * drat: double
#> * wt: double
#> * qsec: double
#> * vs: integer
#> * am: integer
#> * gear: integer
#> * carb: integer
#> * default: guess

## Oh actually cylinders should be a factor
spec$cols$cyl <- col_factor(c("4", "6", "8"))
spec
#> <col_spec>
#> * mpg: double
#> * cyl: factor
#> * disp: double
#> * hp: integer
#> * drat: double
#> * wt: double
#> * qsec: double
#> * vs: integer
#> * am: integer
#> * gear: integer
#> * carb: integer
#> * default: guess

data <- read_csv("mtcars.csv", col_types = spec)
data
#> <tibble [32 x 11]>
#>      mpg    cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>    <dbl> <fctr> <dbl> <int> <dbl> <dbl> <dbl> <int> <int> <int> <int>
#> 1   21.0      6 160.0   110  3.90 2.620 16.46     0     1     4     4
#> 2   21.0      6 160.0   110  3.90 2.875 17.02     0     1     4     4
#> 3   22.8      4 108.0    93  3.85 2.320 18.61     1     1     4     1
#> 4   21.4      6 258.0   110  3.08 3.215 19.44     1     0     3     1
#> 5   18.7      8 360.0   175  3.15 3.440 17.02     0     0     3     2
#> 6   18.1      6 225.0   105  2.76 3.460 20.22     1     0     3     1
#> 7   14.3      8 360.0   245  3.21 3.570 15.84     0     0     3     4
#> 8   24.4      4 146.7    62  3.69 3.190 20.00     1     0     4     2
#> 9   22.8      4 140.8    95  3.92 3.150 22.90     1     0     4     2
#> 10  19.2      6 167.6   123  3.92 3.440 18.30     1     0     4     4
#> ... with 22 more rows

# alternatively can use the spec as a character
cat(as.character(spec))
#> cols(
#>   col_double(),
#>   col_factor(levels = c("4", "6", "8"), ordered = FALSE),
#>   col_double(),
#>   col_integer(),
#>   col_double(),
#>   col_double(),
#>   col_double(),
#>   col_integer(),
#>   col_integer(),
#>   col_integer(),
#>   col_integer())

data2 <- read_csv("mtcars.csv",
  col_types = cols(
    col_double(),
    col_factor(levels = c("4", "6", "8"), ordered = FALSE),
    col_double(),
    col_integer(),
    col_double(),
    col_double(),
    col_double(),
    col_integer(),
    col_integer(),
    col_integer(),
    col_integer()))

all.equal(data, data2)
#> [1] TRUE
```
We can do separate `spec_*` functions, that was actually my first prototype of this.

I went with the argument mainly to decrease the maintenance burden, since you don't have to replicate (and keep in sync) the setup logic in two different places.

I also think the argument is slightly more discoverable than a separate function, but that is largely subjective.

hadley:
I think most people will prefer the manipulating the character strings, but I do like how this gives you both interfaces.

How hard would it be generate a named list? That would make it easier to work with, and then if you wanted to only read a few cols, you could delete the lines you don't want and switch to `cols_only()`.

Maybe the default could be to print it out whenever any columns are guessed? That nice because it tells you exactly how readr is reading it in, and if you're loading a big file and spot a problem you can ctrl + break, fix the problem and start again?

I really dislike arguments that change the return type of a function. But I agree that having `spec_csv()`, `spec_tsv()` etc is going to add quite a bit of code duplication. But there are only 5 functions we'd need to do it for. But it is a frequent source of copy and paste errors. I wonder if we could have a tokenizer helper that reached into the parent environment to pluck out the arguments it needed?

jimhester:
Adding column names is easy, I should have done that originally.

The main issue with printing columns by default is I think it would get unwieldy when you are reading a file with 100+ columns.

That is a good point about the return type polymorphism. I will try some ideas to avoid too much duplication.

hadley:
We could only print (say) the first twenty names by default.

Another option would be to return the spec as an attribute (like problems), but then of course you'd need to wait until you'd parsed the whole thing successfully.

jennybc:
It seems like this is about #314, as much or more than #304 or even #237? BTW this will be so useful.

jimhester:
@jennybc Sorry 304 was a typo, should be #314 as you said!

jimhester:
All the `read_*()` functions include an attribute with the col_spec now, which can be retrieved with `spec()`. There are also `spec_*()` functions for each `read_*()` function.

I also added a `guess_max` parameter to allow users to specify exactly how many lines they wanted to use when guessing, and changed the default value of `n_max` from `-1` to `Inf`, which allows an easy definition of the guess_max default.

The spec functions are all defined to call the read functions with n_max = 0, so they only end up reading the file once.

``` r
write_csv(mtcars, "mtcars.csv")
data <- read_csv("mtcars.csv")
#> cols(
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_integer(),
#>   am = col_integer(),
#>   gear = col_integer(),
#>   carb = col_integer())
data
#> # A tibble: 32 x 11
#>      mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>    <dbl> <int> <dbl> <int> <dbl> <dbl> <dbl> <int> <int> <int> <int>
#> 1   21.0     6 160.0   110  3.90 2.620 16.46     0     1     4     4
#> 2   21.0     6 160.0   110  3.90 2.875 17.02     0     1     4     4
#> 3   22.8     4 108.0    93  3.85 2.320 18.61     1     1     4     1
#> 4   21.4     6 258.0   110  3.08 3.215 19.44     1     0     3     1
#> 5   18.7     8 360.0   175  3.15 3.440 17.02     0     0     3     2
#> 6   18.1     6 225.0   105  2.76 3.460 20.22     1     0     3     1
#> 7   14.3     8 360.0   245  3.21 3.570 15.84     0     0     3     4
#> 8   24.4     4 146.7    62  3.69 3.190 20.00     1     0     4     2
#> 9   22.8     4 140.8    95  3.92 3.150 22.90     1     0     4     2
#> 10  19.2     6 167.6   123  3.92 3.440 18.30     1     0     4     4
#> ... with 22 more rows

# Every table returned has a spec attribute
s <- spec(data)
s
#> cols(
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_integer(),
#>   am = col_integer(),
#>   gear = col_integer(),
#>   carb = col_integer())

# Alternatively you can use a spec function instead, which will only read the
# first 1000 rows (user configurable with guess_max)
s <- spec_csv("mtcars.csv")
#> cols(
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_integer(),
#>   am = col_integer(),
#>   gear = col_integer(),
#>   carb = col_integer())
s
#> cols(
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_integer(),
#>   am = col_integer(),
#>   gear = col_integer(),
#>   carb = col_integer())

# If the spec has a default of skip then uses cols_only
s$default <- col_skip()
s
#> cols_only(
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_integer(),
#>   am = col_integer(),
#>   gear = col_integer(),
#>   carb = col_integer())

# Otherwise set the default to the proper type
s$default <- col_character()
s
#> cols(.default = col_character(),
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_integer(),
#>   am = col_integer(),
#>   gear = col_integer(),
#>   carb = col_integer())

# The print method takes a n parameter to return only that number of columns
print(s, n=5)
#> cols(.default = col_character(),
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double()
#>   # ... with 6 more columns
#> )

# When reading this is set to 20 by default, set options("readr.num_columns" = x) to change
options("readr.num_columns" = 5)
data <- read_csv("mtcars.csv")
#> cols(
#>   mpg = col_double(),
#>   cyl = col_integer(),
#>   disp = col_double(),
#>   hp = col_integer(),
#>   drat = col_double()
#>   # ... with 6 more columns
#> )
```

hadley:
I like it!  But I think it's sufficiently complex and useful that it probably need to go in the column types vignette.

@earino would love your thoughts

earino:
Cool! A few thoughts:

1. I think this is a better version of the thing I was imagining. 
2. If I want to share spec objects between R processes, is the idea that I save them out as an rda and pass them around in that fashion?
3. The examples didn't have dates in them, does it work for that as well?

hadley:
@earino you could either save as RDS or print and copy and paste into your code.

jimhester:
Ok just added some tests, documentation and news for this, should be ready unless you spot something that needs to change.

hadley:
One last thing - I realised it would be useful to reference `spec()` when there are any problems. Could you tweak the message?

Otherwise looks good to merge.

jimhester:
Ok I added a line about using spec().

``` r
read_csv("a,b,c\nd,e,3", col_types='iii')
#> Warning: 2 parsing failures.
#> See spec(...) for column specifications used.
#> row col   expected actual
#>   1   a an integer      d
#>   1   b an integer      e
#> # A tibble: 1 x 3
#>       a     b     c
#>   <int> <int> <int>
#> 1    NA    NA     3
```

hadley:
Perfect!

PR #439, opened 2016-06-17
Revert "Check for BOM when converting"

jimhester:
Reverts hadley/readr#436


jimhester:
Something is going wonky with the boost change, reverting till I figure out the fix.

PR #448, opened 2016-07-06
write_excel_csv() to write a UTF-8 BOM

jimhester:
I had to replicate some logic in `write_excel_csv()` from `write_delim()` because I didn't want to expose an extra parameter in `write_delim()` for writing a byte order mark.


hadley:
Seems like a reasonable trade off - LGTM

PR #452, opened 2016-07-06
read_fwf() on an empty file returns an empty tibble

### LINKED ISSUE #441: read_fwf on empty file throws error
Hi, 

read_fwf on an empty file throws this (Windows-specific I think) error message:
`Error: Cannot read file xxx: The volume for a file has been externally altered so that the opened file is no longer valid.`

Note that this is a new behaviour: before I installed the latest `readr` version, the error was simply `Cannot read file xxx`.

I have the latest `readr` from github (re-installed minutes ago) and I work with R version 3.2.3 on Windows 7.

To reproduce:

    library("readr")
    # create an empty file in working dir
    write_csv(data.frame(), "emptyFile")
    # this line fails
    read_fwf("emptyFile", col_positions=fwf_widths(1))

This is similar to [this issue](https://github.com/hadley/readr/issues/356), which has been solved by returning empty structures when the file was empty. I suggest to extend the fix to `read_fwf`.

jimhester:
Fixes #441

### DIFF
--- a/R/read_fwf.R
+++ b/R/read_fwf.R
@@ -31,6 +31,11 @@ read_fwf <- function(file, col_positions, col_types = NULL,
                      locale = default_locale(), na = c("", "NA"),
                      skip = 0, n_max = Inf, guess_max = min(n_max, 1000), progress = interactive()) {
   ds <- datasource(file, skip = skip)
+
+  if (inherits(ds, "source_file") && empty_file(file)) {
+    return(tibble::data_frame())
+  }
+
   tokenizer <- tokenizer_fwf(col_positions$begin, col_positions$end, na = na)
 
   spec <- col_spec_standardise(
--- a/R/source.R
+++ b/R/source.R
@@ -158,3 +158,7 @@ zipfile <- function(path, open = "r") {
 
   unz(path, file, open = open)
 }
+
+empty_file <- function(x) {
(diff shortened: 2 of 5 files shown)


R/source.R
```diff
@@ -158,3 +158,7 @@ zipfile <- function(path, open = "r") {
 
   unz(path, file, open = open)
 }
+
+empty_file <- function(x) {
+  return(is.character(x) && file.exists(x) && file.info(x, extra_cols = FALSE)$size == 0)
```
hadley: Drop the return?

R/read_fwf.R
```diff
@@ -31,6 +31,11 @@ read_fwf <- function(file, col_positions, col_types = NULL,
                      locale = default_locale(), na = c("", "NA"),
                      skip = 0, n_max = Inf, guess_max = min(n_max, 1000), progress = interactive()) {
   ds <- datasource(file, skip = skip)
+
+  if (inherits(ds, "source_file") && empty_file(ds[[1]])) {
```
hadley: Don't you normally do the check on the `file`?

jimhester: Yeah, it ends up being the same thing as `ds[[1]]`, but it is probably more clean to use file here, thanks.

R/source.R
jimhester: Yes, not sure why I had it to be honest...

PR #456, opened 2016-07-07
Properly ignore embedded nuls.

### LINKED ISSUE #338: Allow read_lines to gracefully skip null values
Unlike `readLines()`, readr::read_lines() doesn't have a `skipNul=TRUE` option. Hence it either chokes on null characters, or strings get truncated at the null character. Could a `skipNul=TRUE` option be implemented?

jimhester:
It looks like the signature of the Token constructor changed, but the
calls in TokenizerLine were not properly updated for the change. It
wasn't causing a compilation error because the hasNull boolean was being
implicitly coerced to an integer, and the 0 was being coerced to a
boolean, (very bad!).

Fixes #338


hadley:
Ooops! LGTM

PR #457, opened 2016-07-07
write_lines() function to write lines to a file

### LINKED ISSUE #302: Need write_file and write_lines
To do reverse of `read_file()` and `read_lines()`  (should always write as utf-8)

jimhester:
Fixes #302 

The issue mentions `write_file()`, but I am not really sure how that would be different than calling `write_lines()` with a 1 element character vector...


hadley:
LGTM.  I was thinking `write_file()` to parallel `read_file()` but you're right that it's probably not necessary (although `write_lines()` will always write a trailing "\n" so it's not perfectly symmetric)

jimhester:
That is true, at least on POSIX systems text files should always end with newlines, so that seems OK.

Issue #321, opened 2015-11-25
Tolerate blank lines ... at least at the end of input?


jennybc:
``` r
library(readr)
input <- "x,y\n1,2\n\n"
read.csv(text = input)
#>   x y
#> 1 1 2
read_csv(input)
#> Warning: 1 parsing failure.
#> row col  expected    actual
#>   2  -- 2 columns 1 columns
#> Source: local data frame [2 x 2]
#> 
#>       x     y
#>   (int) (int)
#> 1     1     2
#> 2    NA    NA
read_csv(trimws(input))
#> Source: local data frame [1 x 2]
#> 
#>       x     y
#>   (int) (int)
#> 1     1     2
```

It seems reasonable to tolerate and ignore blank lines ... at least at the end of the input.

hadley:
Yeah, I think that's totally reasonable (and for empty lines anywhere, not just at the end). It's just a little tricky to implement because you need to think more about what line numbers to report in error messages etc - the original line numbers, or the row numbers in the generated data frame.

PR #459, opened 2016-07-08
Comment parsing bug when there is a line with nothing but comments

### LINKED ISSUE #381: Skip commented lines
I expected the `read.csv()` behaviour (TBH I'm surprised about the lack of `NA`s there).

``` r
library(readr)

mat <- matrix(
  c(paste0("thing", 1:3),
    "A2", "B2", "C2",
    "A3", "#B3", "C3",
    "#A4", "B4", "C4",
    "A5", "*", "C5"),
  byrow = TRUE, ncol = 3)
row_txt <- apply(mat, 1, paste, collapse = ",")
txt <- paste(row_txt, collapse = "\n")

read_csv(txt)
#> Source: local data frame [4 x 3]
#> 
#>   thing1 thing2 thing3
#>    (chr)  (chr)  (chr)
#> 1     A2     B2     C2
#> 2     A3    #B3     C3
#> 3    #A4     B4     C4
#> 4     A5      *     C5
read.csv(text = txt)
#>   thing1 thing2 thing3
#> 1     A2     B2     C2
#> 2     A3    #B3     C3
#> 3    #A4     B4     C4
#> 4     A5      *     C5

read_csv(txt, comment = "#")
#> Warning: 1 parsing failure.
#> row col  expected    actual
#>   2  -- 3 columns 4 columns
#> Source: local data frame [2 x 3]
#> 
#>   thing1 thing2 thing3
#>    (chr)  (chr)  (chr)
#> 1     A2     B2     C2
#> 2     A3     A5      *
read.csv(text = txt, comment.char = "#")
#>   thing1 thing2 thing3
#> 1     A2     B2     C2
#> 2     A3              
#> 3     A5      *     C5
```

jimhester:
Fixes #381

### DIFF
--- a/src/TokenizerDelim.cpp
+++ b/src/TokenizerDelim.cpp
@@ -73,14 +73,20 @@ Token TokenizerDelim::nextToken() {
     switch(state_) {
     case STATE_DELIM:
       if (*cur_ == '\r' || *cur_ == '\n') {
-        newRecord();
         advanceForLF(&cur_, end_);
-        return emptyToken(row, col);
+        if (col_ == 0) {
+          advanceForLF(&cur_, end_);
+          line_++; line++;
+          token_begin = cur_ + 1;
+          break;
+        }
+        newRecord();
+        return emptyToken(row, col, line);
       } else if (isComment(cur_)) {
         state_ = STATE_COMMENT;
       } else if (*cur_ == delim_) {
         newField();
-        return emptyToken(row, col);
+        return emptyToken(row, col, line);
       } else if (*cur_ == quote_) {
         state_ = STATE_STRING;
       } else if (escapeBackslash_ && *cur_ == '\\') {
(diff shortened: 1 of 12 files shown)


src/TokenizerDelim.cpp
```diff
@@ -160,9 +161,17 @@ Token TokenizerDelim::nextToken() {
 
     case STATE_COMMENT:
       if (*cur_ == '\r' || *cur_ == '\n') {
+
+        // If we have read at least one record on the current row go to the
+        // next row, line, otherwise just ignore the line.
+        if (col_ > 0) {
+          newRecord();
+          ++row;
```
hadley: I think we always need to increment `row` otherwise the error messages will be off after lines that contain only coments?

jimhester: That is true, but then you get all NAs for commented out lines in the output, which differs from the current convention (and that of `read.csv()`).

hadley: So maybe we need a second parameter? Lines vs. rows?

jimhester: Yeah maybe that would work best.

hadley: I'm having vague flashbacks that this is why I never fixed this earlier ;) Probaby interacts with skip too

hadley:
Probably same problem in #321

jimhester:
An alternative to tracking rows and lines separately would be to output a row of NAs for every blank line (including commented out lines). Benefits would be simple implementation and consistency between input and output numbers and I imagine there are times where users would want to keep blank lines. Downside is `read.csv()` drops blanks by default.

hadley:
I think we really do need to drop missing lines by default

jimhester:
Ok I will work on it, do we want an option to keep them like `read.csv()`?

hadley:
It doesn't seem that useful to me, so I'd say no (we can always add later if people really want it)

PR #463, opened 2016-07-08
Make defaults consistent

hadley:
Need to either consistently use flexible parser, or format specified in locale
or provide someway to select between them


hadley:
One option would be to provide an extension to the format strings that specify the flexible date/time parsers

PR #468, opened 2016-07-11
Add condense option to print.col_spec

### LINKED ISSUE #466: Truncate option for print.col_spec
Imagine you're reading a csv file with 1000 columns, of which 990 are doubles. It would be useful if `print.col_spec()` could automatically generate a spec with `default = col_double()`, and then only explicitly label the columns that were different from that default.

jimhester:
Not sure if we want to export `condense_spec()`, I could see it being useful though, so we may.

Fixes #466


tests/testthat/test-col-spec.R
```diff
@@ -146,3 +146,18 @@ test_that("print(col_spec) with n == 0 prints nothing", {
...
+  expect_output(print(out, condense = TRUE),
+    regex_escape(
+"cols(.default = col_integer(),
+  d = col_character()
+)"))
+
+  out <- col_spec_standardise("a,b,c,d\n1,2,3,4")
+  expect_output(print(out, condense = TRUE),
+    regex_escape(
+"cols(.default = col_integer()
```
hadley: I think it might be easier to consistently put each argument on its own line

R/col_types.R
```diff
@@ -82,25 +82,43 @@ as.col_spec.default <- function(x) {
...
+  most_common <- names(counts)[counts == max(counts)][[1]]
+
+  x$default <- x$cols[types == most_common][[1]]
+  x$cols <- x$cols[types != most_common]
+  x
+}
+
 #' @export
-format.col_spec <- function(x, n = Inf, ...) {
+format.col_spec <- function(x, n = Inf, condense = FALSE, ...) {
```
hadley: I think it should be condensed automatically in the auto-printing in `read_delimited()`

hadley:
I agree that it's probably useful enough to export

R/col_types.R
hadley: But if we do it automatically we probably need some threshold so you can still see the column names when there are only a few. Maybe use the same threshold as for the ellipses?

R/col_types.R
```diff
@@ -130,11 +153,22 @@ format.col_spec <- function(x, n = Inf, ...) {
...
+#' \code{\link{condense_spec}} takes a spec object and condenses its definition
+#' by setting the default column type to the most frequent type and only
+#' listing columns with a different type.
 #'
 #' @param x The data frame object to extract from
-#' @return The col_spec object used for that data frame.
+#' @return A col_spec object.
 #' @export
+#' @examples
+#' df <- read_csv(system.file("mtcars.csv", package = "readr"))
```
hadley: Can you please use the new `readr_example()`?

jimhester: Sure thing, my system.file call was wrong anyway ;)

jimhester:
- The echo'd specs are now condensed by default when the columns are greater than `n` (default 20).
- The default argument is on the next line like all other arguments
- Uses `readr_example()`
- `condense_spec()` is exported, documented along with `spec()`. (do we like the name?)

R/col_types.R
```diff
@@ -82,30 +82,50 @@ as.col_spec.default <- function(x) {
...
+  counts <- table(types)
+  most_common <- names(counts)[counts == max(counts)][[1]]
+
+  x$default <- x$cols[types == most_common][[1]]
+  x$cols <- x$cols[types != most_common]
+  x
+}
+
+#' @export
+format.col_spec <- function(x, n = Inf, condense = FALSE, ...) {
```
hadley: I'm not sure about this logic.  Maybe it should be `condense = NULL`, then `condense <- condense %||% length(x$cols) > n` ?

hadley: Oh but that would make the logic below more complicated

jimhester: Ok that works for me

jimhester: I think your suggestion is better, previously there was no way to explicitly force the output to be condensed (other than calling condense_spec()) directly.

hadley:
LGTM - I'm fine with `condense_spec()` as a name (although maybe it should be `spec_condense`?

hadley:
LGTM

Issue #472, opened 2016-07-13
Refactor read_fwf to first parse into lines


hadley:
Then parse into fields. Would make logic for handling raggedness much easier

PR #476, opened 2016-07-14
Allow skipping lines in fwf based on a comment string

### LINKED ISSUE #334: Implement comment skipping for read_fwf()
I have fixed width text files with headers, the size of which (the amount of lines to skip) varies greatly from file to file. I could first read a file and figure out the amount of lines to skip before reloading it with the skip value set, but it feels clumsy and it would be nice to be able to do that in one go. Headers/Comments being often identifiable by the first few characters (in my case, the line starts by the letter H).
It would also be nice to have the possibility to save those skipped lines (the whole header) in a variable or a file at the same time as one may need to parse it further to collect some information from it.

jimhester:
Fixes #334


src/TokenizerFwf.cpp
```diff
@@ -6,15 +6,26 @@ using namespace Rcpp;
...
   std::vector<bool> is_white;
 
   size_t row = 0, col = 0;
+  bool hasComment = comment.length() > 0;
   for (SourceIterator cur = begin; cur != end; ++cur) {
     if (row > n)
       break;
 
+    if (col == 0 && hasComment) {
+      boost::iterator_range<const char*> haystack(cur, end);
```
hadley: Can't you pass `starts_with` a `std::string()` directly?

src/TokenizerFwf.cpp
```diff
@@ -119,6 +132,19 @@ Token TokenizerFwf::nextToken() {
   if (!moreTokens_)
     return Token(TOKEN_EOF, 0, 0);
 
+  // Check for comments only at start of line
+  if (col_ == 0 && isComment(cur_)) {
```
hadley: I think this will work in 99% of cases, but will fail rarely because of the crappy design of the whole parser - I think there are two other places where you need to do the same check. I'd say to not worry about that for now, as it's going to be rare, and will be more easily fixed as part of a bigger refactoring (#472)

src/TokenizerFwf.cpp
jimhester: You can (I am doing so here with the comment parameter), but we don't have a string for the source in this function (unless I am missing something).

hadley: Oh duhr, got it

hadley:
LGTM

PR #478, opened 2016-07-14
Implement write_file and write_file_raw

### LINKED ISSUE #474: write_file()
I think we do need a `write_file()` - I now understand how it should differ from `write_lines()`:

* it should throw an error with a character input of length > 1
* it should also accept a raw vector
* it should not add a trailing nl

It could be documented with `write_lines()` though.

jimhester:
Fixes #474


hadley:
I was thinking that this would be one function that automatically picked raw/character writing based on the vector type

jimhester:
I actually did that first, but there are already separate `read_file()` and `read_file_raw()` functions, so I split the write_file() versions to be consistent.

Happy to change it back though.

hadley:
I think this minor asymmetry is necessary because you can impute the types of inputs, but not of outputs.

hadley:
(i.e. please just have one write function)

jimhester:
Yeah that makes sense, ok I will change it.

jimhester:
Ok now there is one function to rule them all.

hadley:
I'll also need one function to bring them all.

PR #480, opened 2016-07-15
Add an na argument to read_lines()

### LINKED ISSUE #479: read_lines probably should have na argument
Since write_lines does, and they should be symmetric.

jimhester:
Fixes #479

### DIFF
--- a/R/read_lines.R
+++ b/R/read_lines.R
@@ -15,7 +15,7 @@
 #' read_lines_raw(readr_example("mtcars.csv"))
 read_lines <- function(file, skip = 0, n_max = -1L,
                        locale = default_locale(),
-                       na = "NA",
+                       na = character(),
                        progress = interactive()) {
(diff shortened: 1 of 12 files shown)


R/read_lines.R
```diff
@@ -15,12 +15,13 @@
 #' read_lines_raw(readr_example("mtcars.csv"))
 read_lines <- function(file, skip = 0, n_max = -1L,
                        locale = default_locale(),
+                       na = "NA",
```
hadley: I do wonder if this should default to something other than "NA". Maybe nothing?

jimhester: Seems reasonable, I'll change it.

jimhester:
I changed the default to `character()`, which allows any line to pass through (including the empty string) by default. `na = ""` will change empty lines to NA.

hadley:
LGTM

PR #483, opened 2016-07-18
Always print .0 after whole number decimals

### LINKED ISSUE #482: Parsing
The function read_csv() returns "Warning: <x> parsing failures" for certain .csv files which were created with write_csv(). The issue seems to occur only for *large*, i.e. >=60 MB, files (unfortunately, github won't let me upload a file > 10MB so I can't share an example case here). A visual examination of csv files written w/ write_csv() look fine, so the issue appears to be with read_csv()'s parsing, not write_csv()'s writing.

jimhester:
This ensures they will be read as decimals in the future.

If we want to go this route the implementation was not too hard to do.

Fixes #482


hadley:
Let's do it!  You can merge once you've added a news bullet.

PR #485, opened 2016-07-19
Callback/Chunked API

### LINKED ISSUE #427: Callback API
Mostly a placeholder for now, but for the next release of readr we need to make sure we have a callback interface so that have a way to handle files that are bigger than memory.

Rough sketch of interface after discussions with @jcheng5

``` R
read_csv_chunked <- function(..., callback) {
  callback <- as_chunk_callback(callback)

  callback$begin()
  on.exit(callback$finally(), add = TRUE)
  pos <- 1

  while (callback$continue() && has_more_rows()) {
    data <- read_more_rows()
    callback$receive(data, pos)
    pos <- pos + nrow(data)
  }

  return(callback$result())
}

as_chunk_callback <- function(x) UseMethod("as_chunk_callback")
as_chunk_callback.function <- function(x) {
  SideEffectChunkCallback$new(callback)
}
as_chunk_callback.R6ClassGenerator <- function(x) {
  as_chunk_callback(x$new())
}
as_chunk_callback.ChunkCallback <- function(x) {
  x
}

# This would be used if the result should be thrown away
SideEffectChunkCallback <- R6Class("SideEffectChunkCallback", "ChunkCallback",
  private = list(
    callbackFunc = NULL,
    cancel = FALSE
  ),
  public = list(
    initialize = function(callbackFunc) {
      check_callback_fun(callbackFunc)
      private$callbackFunc <- callbackFunc
    },
    receive = function(data, index) {
      result <- private$callbackFunc(data, index)
      private$cancel <- identical(result, FALSE)
    },
    continue = function() {
      !private$cancel
    }
  )
)

# Used if the result of each chunk should be combined
# at the end
DataFrameCallback <- R6::R6Class("DataFrameCallback", "ChunkCallback",
  private = list(
    callbackFunc = NULL,
    results = list()
  ),
  public = list(
    initialize = function(callbackFunc) {
      private$callbackFunc <- callbackFunc
    },
    receive = function(data, index) {
      result <- private$callbackFunc(data, index)
      private$results <- c(private$results, list(result))
    },
    finally = function() {
      dplyr::bind_rows(private$results)
    }
  )
)
check_callback_fun <- function(x) {
  n_args <- length(formals(x))
  if (n_args < 2) {
    stop("`callbackFunc` must have two or more arguments", call. = FALSE)
  }
}
```

jimhester:
Currently only `read_lines_chunked()` is implemented.

For the first pass this is simply maintaining external pointers to the tokenizer and source in an R list and passing them to `read_lines_chunked_()`, as I thought it would be easier to get things prototyped faster that way.

Fixes #427


R/callback.R
```diff
@@ -0,0 +1,105 @@
...
+  as_chunk_callback(x$new())
+}
+as_chunk_callback.ChunkCallback <- function(x) {
+  x
+}
+
+ChunkCallback <- R6::R6Class("ChunkCallback",
+  private = list(
+    callback = NULL,
+    cancel = FALSE
```
hadley: I think `cancel` should only be in `SideEffectChunkCallback `

hadley:
@jcheng5 can you take a look too please?

jimhester:
I think there will be too much code duplication in the `read_token_` case to use the previous approach. de23009 uses the same `read_line_()` function as the non-chunked version and calls the R functions if needed. I think the same approach will work well with `read_token_`/ read_delim.

src/read.cpp
```diff
@@ -38,15 +39,27 @@ CharacterVector read_lines_(List sourceSpec, List locale_, std::vector<std::stri
...
     if (progress && (i + 1) % 25000 == 0)
       progressBar.show(tokenizer.progress());
 
-    if (i >= n) {
+    if (chunked && i >= chunk_size) {
+      if (!as<Function>(callback["continue"])()) {
+        break;
+      }
+      out = Rf_xlengthgets(out, i);
+      as<Function>(callback["receive"])(out, pos);
```
hadley: Maybe make little wrapper like/

```
Function R6method(Environment env, std::string method) {
   return as<Function>(env[method]);
}
```

?

src/read.cpp
```diff
@@ -38,15 +39,27 @@ CharacterVector read_lines_(List sourceSpec, List locale_, std::vector<std::stri
...
 
   R_len_t i = 0;
+  R_len_t pos = i + 1;
   for (Token t = tokenizer.nextToken(); t.type() != TOKEN_EOF; t = tokenizer.nextToken()) {
     if (progress && (i + 1) % 25000 == 0)
       progressBar.show(tokenizer.progress());
 
-    if (i >= n) {
+    if (chunked && i >= chunk_size) {
+      if (!as<Function>(callback["continue"])()) {
```
hadley: Shouldn't `receive()` get called before `continue()`?

src/read.cpp
```diff
@@ -38,15 +39,27 @@ CharacterVector read_lines_(List sourceSpec, List locale_, std::vector<std::stri
   LocaleInfo locale(locale_);
   Progress progressBar;
 
+  bool chunked = chunk_size != -1;
+
   R_len_t n = (n_max < 0) ? 10000 : n_max;
+
   CharacterVector out(n);
```
hadley: Shouldn't this be affected by the chunk size?

src/read.cpp
jimhester: Seems like a good idea

src/read.cpp
jimhester: The conditional to set n should probably take chunk_size into account, just need to think about the correct logic a little.

src/read.cpp
jimhester: If we break first we will still call receive at the end outside the loop (https://github.com/hadley/readr/pull/485/files#diff-4d4cc0b349b9719386d382347a55305dR82).

R/callback.R
```diff
@@ -0,0 +1,83 @@
+read_lines_chunked <- function(file, callback, skip = 0, n_max = -1L, chunk_size = 10000,
+                       locale = default_locale(), na = character(), progress = interactive()) {
+  if (empty_file(file)) {
+    return(character())
+  }
+  ds <- datasource(file, skip = skip)
+
+  callback <- as_chunk_callback(callback)
+  on.exit(callback$finally(), add = TRUE)
+  pos <- 1
```
jcheng5: Is this a stray `pos`?

src/read.cpp
jcheng5: Seems weird to me too. It seems especially weird that `receive()` would get called again even after `continue()` returns `FALSE`. I was imagining the equivalent of:

```
while (callback.continue()) {
  callback.receive(readNextChunk())
}
```

jcheng5:
I only started to look at your code and now I have to run for a little bit, but I wanted to ask whether `read_lines_` could be simplified further by making the non-chunked codepath just a special case of the chunked codepath? i.e. `read_lines_` would only know how to do chunked, and you could build the non-chunked codepath on top of that.

src/read.cpp
```diff
@@ -38,15 +39,27 @@ CharacterVector read_lines_(List sourceSpec, List locale_, std::vector<std::stri
...
   for (Token t = tokenizer.nextToken(); t.type() != TOKEN_EOF; t = tokenizer.nextToken()) {
     if (progress && (i + 1) % 25000 == 0)
       progressBar.show(tokenizer.progress());
 
-    if (i >= n) {
+    if (chunked && i >= chunk_size) {
+      if (!as<Function>(callback["continue"])()) {
+        break;
+      }
+      out = Rf_xlengthgets(out, i);
```
jcheng5: Curious why we use these C APIs instead of the more obvious `CharacterVector.push_back()`? Are the C APIs faster?

src/read.cpp
```diff
@@ -38,15 +39,27 @@ CharacterVector read_lines_(List sourceSpec, List locale_, std::vector<std::stri
...
+      as<Function>(callback["receive"])(out, pos);
+      pos = i + 1;
+      i = 0;
+    } else if (!chunked && i >= n) {
       if (n_max < 0) {
         // Estimate rows in full dataset
         n = (i / tokenizer.progress().first) * 1.2;
```
jcheng5: This is a comment for line 62/75 below (`++i`). Shouldn't that be inside the conditional just above it, that assigns `out[i]`? Or better yet, `out[i++] = ...`. Otherwise, it seems like we're leaving garbage data in `out` everywhere we see a token that should be ignored...?

R/callback.R
```diff
@@ -0,0 +1,83 @@
...
+ChunkCallback <- R6::R6Class("ChunkCallback",
+  private = list(
+    callback = NULL
+  ),
+  public = list(
+    initialize = function() NULL,
+    receive = function() NULL,
+    continue = function() TRUE,
+    finally = function() NULL,
+    result = function() NULL
```
jcheng5: Use correct signatures for these functions?

jcheng5:
^ And/or, how about extracting out a function to "retrieve the next up-to-_n_ number of tokens".

```r
CharacterVector out(n);
var pos = 1;
while (callback.continue() && getMoreTokens(&out, n)) {
  if (chunked) {
    callback.receive(out, pos);
    pos += out.length;
    out.clear();
  }
}
```

(This does result in one more call to `callback.continue()` than is strictly necessary, which we could remove by having `getMoreTokens` also set a bool indicating whether the stream of tokens is done)

I hope I'm not bikeshedding too much, but I feel like this is a pretty critical bit of the code, so worth polishing until it's not just correct but "obviously correct".

R/callback.R
jimhester: Yes thanks, missed removing it when I refactored this.

src/read.cpp
jimhester: `CharacterVector.push_back()` does a full copy on every call (the same behavior as `txt <- "abc";txt <- append(txt, "xyz")` (http://stackoverflow.com/a/13783044/2055486).

As to why readr uses the C APIs and grows them by hand instead of `std::vector<std::string>` @hadley would have to say, I imagine encodings has something to do with it.

src/read.cpp
jimhester: That is a good point actually and looks like a bug in the current implementation to me. `out[i++]` seems like the right fix. I am not sure if this is actually a problem in practice, it doesn't look like`TokenizerLine()` can actually produce a `TOKEN_EMPTY` value. Regardless this could easily change in the future, so it should be changed.

jimhester:
I like the idea of https://github.com/hadley/readr/pull/485#issuecomment-233779510, I find the logic of this function (and `read_tokens_` a little difficult to discern without close examination) and agree making them 'obviously correct' seems like a good idea.

src/read.cpp
hadley: It's hard to know exactly how the C++ wrappers interact with the protection stack. I often prefer the raw C API because it's easier to know exactly what's happening.

The goal of using R data structures instead of C++ data structures is to avoid a copy

jimhester:
I tried to make the logic a little cleaner in fff0d06, mainly by using `std::vector<SEXP>` and `push_back()` instead of the C API, which removes the need to handle growing the vector ourselves.

I did a prototype with `getMoreTokens()`, the main issue is you need to pass the tokenizer, locale and progressBar to the function for every call, so it obscures the intent a bit.

It might be worth defining a simple class that takes a sourceSpec, tokenizer and locale and has a `getMoreTokens()` function. If I went that route I would like it to work for the `read_token_` case as well, which is more complicated.

src/Source.cpp
```diff
@@ -7,20 +7,24 @@ using namespace Rcpp;
 #include "SourceRaw.h"
 
 SourcePtr Source::create(List spec) {
+  return SourcePtr(Source::createDumbPtr(spec));
```
hadley: This can be reverted now, right?

jimhester: Yeah

src/read.cpp
```diff
@@ -38,39 +49,40 @@ CharacterVector read_lines_(List sourceSpec, List locale_, std::vector<std::stri
...
 
-  R_len_t n = (n_max < 0) ? 10000 : n_max;
-  CharacterVector out(n);
+  bool chunked = chunk_size != -1;
 
-  R_len_t i = 0;
-  for (Token t = tokenizer.nextToken(); t.type() != TOKEN_EOF; t = tokenizer.nextToken()) {
-    if (progress && (i + 1) % 25000 == 0)
-      progressBar.show(tokenizer.progress());
+  std::vector<SEXP> out;
```
hadley: This seems really dangerous to me - who is responsible for protecting the SEXPs against gc?

jimhester: That is true, I switched it back to the previous implementation.

jimhester:
Closing this in favor of a new attempt #490

PR #489, opened 2016-07-21
Refactor read functions

jimhester:
First pass at this is now working.

I may write a couple more constructors for the Reader class, so you don't have to do as much object instantiation up front.

It should be pretty straightforward to write chunked functions using this as well.

### DIFF
--- a/src/Progress.h
+++ b/src/Progress.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <time.h>
 
 inline int now() {
--- a/src/Progress.h
+++ b/src/Progress.h
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef FASTREAD_PROGRESS_H_
+#define FASTREAD_PROGRESS_H_
 
 #include <time.h>
 
@@ -79,3 +80,4 @@ public:
 
 };
 
+#endif
--- a/src/Reader.cpp
+++ b/src/Reader.cpp
@@ -1,17 +1,29 @@
 #include "Reader.h"
 
 Reader::Reader(SourcePtr source, TokenizerPtr tokenizer, std::vector<CollectorPtr> collectors,
-                    CharacterVector colNames, LocaleInfo* locale,
-                    int progress) :
+        bool progress, LocaleInfo* locale, CharacterVector colNames) :
   source_(source),
   tokenizer_(tokenizer),
   collectors_(collectors),
-  locale_(locale),
-  progress_(progress) {
+  progress_(progress),
+  locale_(locale) {
+    init(colNames);
+}
 
+Reader::Reader(SourcePtr source, TokenizerPtr tokenizer,
+    CollectorPtr collector, bool progress, LocaleInfo* locale, CharacterVector colNames) :
+  source_(source),
+  tokenizer_(tokenizer),
+  progress_(progress),
+  locale_(locale) {
+
+  collectors_.push_back(collector);
+  init(colNames);
+}
+
+void Reader::init(CharacterVector colNames) {
   tokenizer_->tokenize(source_->begin(), source_->end());
   tokenizer_->setWarnings(&warnings_);
-
 
   // Work out which output columns we are keeping and set the locale for each collectors
   size_t p = collectors_.size();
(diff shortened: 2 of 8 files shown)


src/Progress.h
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
hadley: According to wikipedia, this isn't supported by Solaris Studio C/C++ :/

src/Reader.cpp
```diff
@@ -0,0 +1,114 @@
...
+  // Work out which output columns we are keeping and set the locale for each collectors
+  size_t p = collectors_.size();
+  for (size_t j = 0; j < p; ++j) {
+    if (!collectors_[j]->skip()) {
+      keptColumns_.push_back(j);
+      collectors_[j]->setWarnings(&warnings_);
+    }
+  }
+
+    if (colNames.size() > 0) {
```
hadley: Is this indented correctly?

src/Progress.h
jimhester: Wow, Ok I will change it to header guards. Solaris strikes again :(

src/Reader.cpp
jimhester: No!

hadley:
Yeah, a constructor that accepts a single collector would be nice.

src/Reader.h
```diff
@@ -0,0 +1,36 @@
...
+    Reader(SourcePtr source, TokenizerPtr tokenizer, std::vector<CollectorPtr> collectors,
+        CharacterVector colNames, LocaleInfo* locale, int progress = 10000);
+
+    RObject readToDataFrame(int lines = -1);
+
+    template <typename T>
+    T readToVector(int lines) {
+      read(lines);
+
+      return as<T>(collectors_[0]->vector());
```
hadley: Does this need to add warnings too?

jimhester: We weren't adding warnings before for read_line or read_raw, but I can do so if you think they would be useful.

src/Reader.cpp
```diff
@@ -0,0 +1,114 @@
...
+  out.attr("names") = outNames_;
+
+  return warnings_.addAsAttribute(out);
+}
+
+int Reader::read(int lines) {
+
+  int n = (lines < 0) ? 1000 : lines;
+
+  collectorsResize(collectors_, n);
```
hadley: Can this now become a method of `Reader`?

src/Reader.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  int last_row = -1, last_col = -1, cells = 0;
+  for (Token t = tokenizer_->nextToken(); t.type() != TOKEN_EOF; t = tokenizer_->nextToken()) {
+    if (progress_ && (cells++) % progressStep_ == 0) {
```
hadley: I don't think `progressStep_` needs to be a variable - it can just be the same for everything

hadley: The check should probably be more like:

```cpp
if ((cells++) % 10000 == 0) {
   if (progress) {
     progressBar_.show(tokenizer_->progress());
   } else {
      Rf_checkForInterrupts()
   }
}

hadley:
The overall refactoring looks good, although I'm no longer super familiar with the internals. I spotted a few smaller things here and there.

src/Reader.h
hadley: Maybe this should power `parse_vector_()` too so we should at least have `readToVectorWithWarnings()`?

src/Reader.cpp
jimhester: Yeah definitely.

src/Reader.cpp
jimhester: I had it as a variable because the value was different between read_lines_ and read_tokens_ (25,000 vs 250,000 IIRC), but I am happy to keep it constant!

jimhester: We already check for user interrupts within the Tokenizers. Maybe (probably) it would be better moved to the Reader class though...

hadley:
Yeah, let's move the interrupt checks to the one place (the reader)

hadley:
Per slack conversation, the sources are currently tightly bound to contiguous data, so for now I think we'll have to keep the interrupt checks where they are.  

LGTM

jimhester:
Ok thanks for reviewing!

PR #490, opened 2016-07-22
Chunked reading for lines and delimited files

jimhester:
This is working for both cases now. It builds on #489.

The main thing missing is more thorough documentation, and possibly tests of all the read_*_chunked functions, not just read_lines and read_csv.


R/read_delim_chunked.R
```diff
@@ -0,0 +1,100 @@
...
+
+utils::globalVariables(c("callback", "chunk_size"))
+
+read_delimited_chunked <- generate_read_delimited_chunked(read_delimited)
+
+#' Read a delimited file by chunks
+#'
+#' @inheritParams read_delim
+#' @param callback A callback function to call on each chunk
+#' @param chunk_size The number of rows to include in each chunk
```
hadley: I think you can `@keywords internal` for this release

hadley:
@jcheng5 want to take another look? @jimhester refactored quite a bit of the internals so that implementing chunking was more straightforward.

R/callback.R
```diff
@@ -0,0 +1,67 @@
...
+  public = list(
+    initialize = function(callback) {
+      private$callback <- callback
+    },
+    receive = function(data, index) {
+      result <- private$callback(data, index)
+      private$results <- c(private$results, list(result))
+    },
+    result = function() {
+      dplyr::bind_rows(private$results)
```
jcheng5: dplyr dependency

src/Reader.cpp
```diff
@@ -57,42 +59,62 @@ RObject Reader::readToDataFrame(int lines) {
...
       checkColumns(last_row, last_col, collectors_.size());
     }
 
-    if (lines >= 0 && t.row() >= lines) {
+    if (lines >= 0 && t_.row() - first_row >= lines) {
       break;
     }
 
-    if (t.row() >= n) {
+    if ((last_row - first_row) > n) {
```
jcheng5: last_row may be -1 here, is that ok?

jimhester: last_row will always be -1 in the first iteration, so it is ok, but still worth adding `last_row != -1` to be explicit.

R/callback.R
jimhester: Wn can change this to the slower
```r
do.call(`rbind`, private$results)
```
Or put dplyr in `Suggests:` and add a `loadNamespace()` call.

I also should write a test using a `DataFrameCallback` to verify it works. :)

hadley: Or we could just not export this one, saving it for the chunked package.

R/callback.R
```diff
@@ -0,0 +1,67 @@
...
+  SideEffectChunkCallback$new(x)
+}
+as_chunk_callback.R6ClassGenerator <- function(x) {
+  as_chunk_callback(x$new())
+}
+as_chunk_callback.ChunkCallback <- function(x) {
+  x
+}
+
+ChunkCallback <- R6::R6Class("ChunkCallback",
```
hadley: Need to export this

R/callback.R
jimhester: I went with rbind for now, I think it is a useful example class, we can always have a more performant/robust one in the future package.

The DataFrameCallback needed a slight tweak to work anyway, you need to define the retrieve variable or else you get a locked namespace error when R tries to create the new variable in the object's environment.

jimhester:
Is there anything else code or API wise we still need to change or any additional documentation you think is needed?

hadley:
LGTM. Any last thoughts @jcheng5?

jcheng5:
One last thought I've been meaning to communicate since last week. I believe the continue callback is called only after the next chunk has been read--ie it protects the receive callback instead of the read, and I think it would make more sense the other way around. 

I don't think it actually matters much for readr, but something to keep in mind as we generalize the pattern to other backends. This especially gets important when you have a stream that cannot be re-read, like sockets, that you might want to continue chunk-reading later.

jimhester:
That is a good point @jcheng5, I re-wrote the boolean so continue is checked before reading (https://github.com/hadley/readr/pull/490/commits/917465e6cf6125e0ddd2bda5635be19114605cf2)

src/read.cpp
```diff
@@ -60,12 +61,16 @@ void read_lines_chunked_(List sourceSpec, List locale_, std::vector<std::string>
...
+    (out = r.readToVector<CharacterVector>(chunkSize)).size();
+
   int pos = 1;
-  while (out.size() > 0 && R6method(callback, "continue")()) {
+  while (moreLines) {
     R6method(callback, "receive")(out, pos);
     pos += out.size();
-    out = r.readToVector<CharacterVector>(chunk_size);
+    moreLines = R6method(callback, "continue")() &&
+      (out = r.readToVector<CharacterVector>(chunkSize)).size();
```
jcheng5: Looks good! But isn't this also just the same as:

```cpp
while (R6method(callback, "continue")() && (out = r.readToVector<CharacterVector>(chunkSize)).size()) {
  R6method(callback, "receive")(out, pos);
  pos += out.size();
}
```

Other than that, LGTM! Nice work!

jimhester: I thought it was a little more clear to have a variable, happy to remove it if you think it is better without though.

jimhester: Actually it makes more sense to remove it and only have the conditional defined once, thank you for the suggestion!

jimhester:
Merged this into master with https://github.com/hadley/readr/commit/cebbfcd51ec70a7f82a29ac4aece156782e5ce70 for future reference.

PR #498, opened 2016-08-04
Properly allow stopping of chunked reading

jimhester:
Previously we were 'checking' the SEXP address, which would always be
non-zero. This meant that there was no value returned by callback that
would halt the reading.

Instead we need to explicitly test the value of the first element of the logical vector.

I unfortunately realized this this morning instead of yesterday, unfortunate timing considering we released yesterday 😞 .

### DIFF
--- a/src/read.cpp
+++ b/src/read.cpp
@@ -47,7 +47,10 @@ Function R6method(Environment env, const std::string& method) {
   return as<Function>(env[method]);
 }
 bool isTrue(SEXP x) {
-  return TYPEOF(x) == LGLSXP && Rf_length(x) == 1 && LOGICAL(x)[0] == TRUE;
+  if (!(TYPEOF(x) == LGLSXP && Rf_length(x) == 1)) {
+    stop("`continue()` must return a length 1 logical vector");
+  }
+  return LOGICAL(x)[0] == TRUE;
 }
 
 // [[Rcpp::export]]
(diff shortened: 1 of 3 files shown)


src/read.cpp
```diff
@@ -46,6 +46,9 @@ CharacterVector read_lines_(List sourceSpec, List locale_,
 Function R6method(Environment env, const std::string& method) {
   return as<Function>(env[method]);
 }
+bool isTrue(SEXP x) {
+  return TYPEOF(x) == LGLSXP && Rf_length(x) == 1 && LOGICAL(x)[0] == TRUE;
```
hadley: I wonder if we should throw an explicit error if the output is the wrong type (i.e. not length 1 logical vector)

hadley:
Doh!  No doubt we'll find a few more small issues like this and can do a point release in a week or two.

src/read.cpp
jimhester: Maybe, our current R6 class tests with [`identical()`](https://github.com/hadley/readr/blob/a714a8ed01c6e8d51dd9bb790db0b6fbc2bf8146/R/callback.R#L70) so the value of `continue()` will always be either `TRUE` or `FALSE`, but I agree it makes sense to be explicit for user defined callback clasess.

hadley:
Or 6 months later :/

PR #502, opened 2016-08-10
Update RCPP Dependency

michaelquinn32:
write_delim.cpp uses vector.at() which was added in RCPP 0.12.0.5.

See change log for https://github.com/RcppCore/Rcpp/commit/e4868b8f3918dc2f9633c25d8169b53b0a22d1ad


hadley:
Thanks!

PR #520, opened 2016-09-13
ListCallback

mmuurr:
`DataFrameCallback` is great for returning tables.
`SideEffectChunkCallback` is great for side effects.

`ListCallback` is the more general version of `DataFrameCallback` which simply returns the chunked results... which is good for processing (e.g. JSON parsing) via `read_lines_chunked`.


hadley:
I don't get why you'd want this - the chunks will have the same columns so you'd be better off with a data frame than a list.  Can you please file an issue with a motivating example?

PR #527, opened 2016-10-03
Make spec inclusion optional

Ironholds:
Including the spec is useful 9/10ths of the time and the 10th - when you want to str() a data frame with a particularly large number of columns - it fills up the console with the spec and leaves the actual data far away in scrollback.

At the same time: useful 9/10ths of the time! So this pull request just adds `include_spec` as a parameter, set to TRUE by default, and leaves it up to the user. Comes with unit tests and everything.


hadley:
You can just do `col_type = cols()`  to suppress the output

Ironholds:
Oh, clever!

PR #533, opened 2016-10-14
read_csv modification to accept unused arguments with documentation updated

vbs1:
Updated documentation to have the new "..." argument to prevent CI warning. 

prevent Httr unused arguments error encountered during parsing.

Possible unused argument error

Error in readr::read_csv(x, locale = readr::locale(encoding = encoding), : 
unused arguments (check.names = FALSE, na.strings = "\N", colClasses = c(NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA, NA))

Possible Solution

Modify read_delim.R line 109 to used ", ..." in function which would allow for unused arguments to be used, and ignored.

Modification:

read_csv <- function(file, col_names = TRUE, col_types = NULL,
quoted_na = TRUE, comment = "", trim_ws = TRUE, skip = 0,
n_max = Inf, guess_max = min(1000, n_max),
progress = interactive(), ...)

Test

While not the best adding ", ..." to read_csv function allowed the parsing of the content, and resolved the unused arguments error. I'm still new to R and open to suggestions.


jimhester:
Can you provide a reproducible example (reprex) of the httr call that causes the error you are trying to fix?

Probably this is better fixed in httr than by ignoring additional arguments in readr.

vbs1:
I'll get an example up today.

Vincent Blain

hadley:
Yes, this doesn't seem like the right fix to me.

vbs1:
Got busy with family. I'll try to have the example up soon. It was the only
fix it could see. But I'm novice in Russia.

Vince

PR #536, opened 2016-10-20
Vignette Readr.Rmd had no Title

cderv:
This vignette had *Vignette Title* as Title. It is better to have an explicit name, especially for the pkgdown website.
As it present output, I call it *Output* not *readr* as the Rmd file.


cderv:
did not pay attention that you were working on it right now. sorry.

PR #543, opened 2016-10-28
fix typo in doc (guess -> guesses)

daattali:
I'm using CRAN roxygen version and when documenting I got over a dozen files with changes, so I thought it'd be safer if I just don't run roxygen.  I can see in DESCRIPTION that you're using roxygen 5.0.1.9000, which can mean any version from Nov 2015 until now, so I prefered to just leave as is if that's ok?


daattali:
(note to self) I'll do the PR mentioned in https://github.com/tidyverse/readr/issues/530 after a response here that tells me what to do with roxygen

Issue #537, opened 2016-10-20
parse_number doesn't parse scientific notation


aaronmcdaid:
`readr::parse_number('3e-5')`  returns `3` where I expected `0.00003`.

Perhaps this isn't quite a bug, I'm not sure what the designed behaviour is. I thought it would read any 'numeric'-type data.

At the least, the documentation could be improved to specify this more clearly, and to guide users towards `col_types="d"` instead of `"n"` when dealing with floating point in functions like `read_delim`.

I guess this is the relevant sentence from the vignette:

  "... it ignores non-numeric prefixes and suffixes"

Is scientific notation a "non-numeric suffix"?

aaronmcdaid:
I've only just realized that this issue has been fixed here. It was an issue in the 1.0.0 on CRAN, but not in the latest code on github. Sorry for not checking earlier!

aaronmcdaid:
Sorry to reopen. Turns out I did the wrong test. Anyway, the issue is still open. I'm looking at the code in `QiParsers.h` and hope to send a pull request soon. I think I can make a fix to include trailing `e` or `E` part of `parse_number`, if it is followed by an `+` or `=` (optional) and one or more digits.

aaronmcdaid:
Pull request with proposed fix, just submitted: https://github.com/tidyverse/readr/pull/545

holstius:
+1. I was bitten by this when handed a CSV file saved by someone else from Excel. A few hundred rows down, there was a value in scientific notation. Everything else in the column was just digits (and ".")! Took hours to figure out what was happening. :-/

hadley:
Why can't you use `parse_double()`?

holstius:
Thanks! didn't know about it. 

Several things seem to be steering folks looking for a quick way forward --- from legacy stuff like extract_numeric --- to parse_number instead of parse_double: autocomplete; deprecation warnings suggesting parse_number but (IIRC) not parse_double; and forum threads. 

The second seems like the most actionable. I'm off grid or would offer fix ideas. CC'ing myself for New Year's follow-up. 

David

PR #545, opened 2016-10-31
issue #537. extend parse_number to allow scientific notation

aaronmcdaid:
[ Relevant issue I reported  https://github.com/tidyverse/readr/issues/537 ]

Originally, `readr::parse_number('3e-5')` returned `3` where I expected `0.00003`. It appears that `parse_number` is designed to ignore any content that is not obviously a number. This is relevant for `col_type='n'` when reading files.

This pull request extends `parse_number` to also apply any scientific notation that is found after the (non-scientific) number. An `e` or `E`, followed (optionally) by `+` or `-`, followed (non-optionally) by one or more digits, is treated as scientific notation.

`3.14e+` and `3.14e` will fail, by design, as it expects one or more digits after the `e`/`e+`/`e-`. If you prefer more relaxed behaviour (accepting, and ignoring, the `e`/`e+`/`e-` here), then you can delete the following line in my pull request

        seenNumber = false; // to 'force' it to expect more digits

### DIFF
--- a/src/QiParsers.h
+++ b/src/QiParsers.h
@@ -32,6 +32,8 @@ enum NumberState {
   STATE_INIT,
   STATE_LHS,
   STATE_RHS,
+  STATE_SCIENTIFIC_NOTATION_INIT,   // e/E +/-
+  STATE_SCIENTIFIC_NOTATION_DIGITS, // the digits
   STATE_FIN
 };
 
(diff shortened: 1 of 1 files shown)


hadley:
As mentioned in that issue, you should use `parse_double()` here

PR #560, opened 2016-12-05
Document %I format, recommend for use with AM/PM

izahn:
Support for `%I` in `parse_datetime` was added in https://github.com/tidyverse/readr/pull/426, but this was not documented. This PR documents this format, recommending using it with AM/PM.


hadley:
LGTM

jimhester:
Thanks!

PR #575, opened 2017-01-13
CLion support files

krlmlr:
No functional changes.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -15,3 +15,5 @@
 ^README\.Rmd$
 ^README-.*\.png$
 ^docs$
+^CMakeLists\.txt$
+^clion-test\.R$
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ bench/*.csv
 inst/doc
 .DS_Store
 .httr-oauth
+/.idea
+/cmake-build-debug
+clion-test.R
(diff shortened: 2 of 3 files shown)


jimhester:
I don't feel editor specific files should be included in the repository. I feel the same way about .Rproj files, but that ship has sailed.

If you don't want these files tracked by git you can add them to [`.git/info/exclude`](http://stackoverflow.com/a/6117232/2055486) rather than adding them to the project `.gitignore`.

krlmlr:
How would that work with `.Rbuildignore`?

jimhester:
You can set it to be untracked [`git update-index --assume-unchanged .Rbuildignore`](http://stackoverflow.com/a/10755704/2055486), so your local changes would not be commited

krlmlr:
Yes, but how can I include the support files for my IDE from `R CMD build`? Arguably I don't need it that often in this repo, but I wonder about the general case.

jimhester:
You would make the changes to `.Rbuildignore` locally to exclude your CLion files, but not commit them and use the above command to tell git to assume the file is unchanged. If you do need to commit a file to `.Rbuildignore` you can use `git update-index --no-assume-unchanged` to turn tracking back on and commit the needed lines.

I have a few support files for You Complete Me for C++ completion as well that I use locally, but I do not include them in the repository or in the built package.

krlmlr:
This is clever, I'll keep that in mind.

PR #577, opened 2017-01-13
Suggest dplyr

krlmlr:
for tibble 1.3.0 compatibility. Also includes revdep check -- no regressions found.


jimhester:
Can you explain why we need dplyr in Suggests for tibble 1.3.0? Including it is going to slow down every travis and appveyor build...

krlmlr:
It's needed because `all.equal.tbl_df()` now forwards to dplyr and is used by one of the readr examples. I can't release tibble to CRAN because of this.

With caching the slowdown will only apply very rarely.

jimhester:
I ended up falling back to using `all.equal.list()` for tbl_df in the tests (which is where the error was occurring with devel tibble) rather than pulling in the dplyr dependency.

krlmlr:
Thanks. Would you like to merge in the revdep check results (with diff script, also part of this PR)?

jimhester:
We are going to be doing some more extensive work on readr for the next couple of weeks, so I am going to have to re-run revdeps afterwards. I don't think it is worth committing your results now. But I will keep the script in mind to possibly add later.

Issue #514, opened 2016-09-02
Support for True/False in col_logical


rgknight:
I can give this a try, following the example of https://github.com/hadley/readr/pull/406/commits/90041386e592575c083f909a890fd6f74f01ba32

Unless this is possible already?

hadley:
Maybe we should just make `col_logical()` case insensitive?

jimhester:
Now fixed (https://github.com/tidyverse/readr/commit/94b50da31104b65368a43d2cb632e9f027b4c651)

PR #580, opened 2017-01-18
Support True False logical #514

rgknight:
I tried implementing a case insensitive comparison but it was beyond me. True / False should get us most of the way there.

### DIFF
--- a/src/Collector.cpp
+++ b/src/Collector.cpp
@@ -265,13 +265,13 @@ void CollectorLogical::setValue(int i, const Token& t) {
       }
       break;
     case 4:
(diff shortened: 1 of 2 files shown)


jimhester:
To do case insensitive matching you can convert the field to lower case and compare with lower case versions of `true` and `false`.

rgknight:
Yes, it was the particulars of C++ that stumped me, having never worked in it before. I was using `boost::iequals()` to make case insensitive comparisons. The tests in `test-parsing-logical` passed but some other ones failed, e.g., I believe `test-parsing` and "spec object attached to read data", which made me worried I would introduce bugs by using a tool I didn't fully understand.

jimhester:
Thanks, R provides a cross platform `strncasecmp()` that can be used in this case for case insensitive comparison.

PR #582, opened 2017-01-21
change progress defaults

jrnold:
Change defaults for displaying the progress bar to be similar to those for dplyr. Currently, the progress bar defaults to display when not interactive. However, it will still display in knitr documents if they are compiled in an interactive session.

This patch defines an unexported function `progress_defaults` which sets checks for an option `readr.show_progress`, whether it is interactive, and whether it is in a knitr document (`knitr.in.progress`). This is equivalent to how `dplyr` defaults for showing a progress bar.


jimhester:
Thanks!

PR #584, opened 2017-01-23
`parse_time()` now correctly handles 12 AM/PM

### LINKED ISSUE #579: parse_time fails if 12 pm
```R
parse_time("12:01 PM")
```

Because `DateTimeParser.hour` always adds 12. It should subtract 12 if hour is already 12.

From @cwickham

jimhester:
Fixes #579

### DIFF
--- a/src/DateTimeParser.h
+++ b/src/DateTimeParser.h
@@ -311,7 +311,24 @@ public:
 
 private:
   int hour() {
-    return hour_ + (amPm_ == 1 ? 12 : 0);
+    if (hour_ == 12) {
+
+      // 12 AM
+      if (amPm_ == 0) {
+        return hour_ - 12;
+      }
+
+      // 12 PM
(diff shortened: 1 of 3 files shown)


src/DateTimeParser.h
```diff
@@ -311,7 +311,10 @@ class DateTimeParser {
 
 private:
   int hour() {
-    return hour_ + (amPm_ == 1 ? 12 : 0);
+    if (hour_ == 12 && amPm_ == 0) {
```
hadley: I feel like this might be simpler if you branched on `hour_ == 12` so you could also simplify the ternary if below.

jimhester: Ok I did so with https://github.com/tidyverse/readr/pull/584/commits/6f93d5bd265735a2348a60ed8d62bad02e0b5da4

Issue #534, opened 2016-10-17
support overlapping fwf columns


gergness:
I sometimes receive fixed with files with specifications that have overlapping columns. readr's current behavior made it difficult to figure out what was going wrong. 

I wish that I was able to read in the data with the original specifications and get the 2 variables as normal, but if that's not possible, perhaps fwf_positions could check for overlaps and throw a better error?

``` r
library(readr)
read_fwf("2015a\n2016b", fwf_positions(c(1, 3, 5), c(4, 4, 5)))
#> Warning: 4 parsing failures.
#> row col  expected    actual
#>   1  X2 2 chars   1        
#>   1  -- 3 columns 2 columns
#>   2  X2 2 chars   1        
#>   2  -- 3 columns 2 columns
#> 
#> # A tibble: 2 × 3
#>      X1    X2    X3
#>   <int> <chr> <chr>
#> 1  2015     a  <NA>
#> 2  2016     b  <NA>
```

lucasmation:
You are right.  Bump!

I am having an embarrassing problem in a package ([microdadosBrasil](https://github.com/lucasmation/microdadosBrasil)) because of the error reading overlapping columns in the original fwf data source. 

I am posting a new example just to be free. 

Now only are the overlaping columns wrong, but also all the following columns have the initial and end positions shifted right, thus making all the data wrong. 


Let's say I have this import dictionary 
```
# name: initial position to final position
# X1: 1 to 2
# X2: 3 to 4
# X3: 3 to 7 (overlaps first two caracters with X2)
# X4: 8 to 9
# X5: 10 to 14
```
And this dataset

```
123456789abcdefg
123456789abcdefg

```
We can import this using
```

'123456789abcdefg\n123456789abcdefg' %>% 
  read_fwf(fwf_positions(start=c(1,3,3,8,10 ), 
                         end=  c(2,4,7,9,13 ) )
  ) 
```
Thsi leads the data being imported with errors. In particular X3 and every column after that are wrong: 

```
# A tibble: 2 × 5
     X1    X2    X3    X4    X5
  <int> <int> <int> <chr> <chr>
1    12    34 56789    ab  cdef
2    12    34 56789    ab  cdef
```

gergness:
In case it is helpful @lucasmation, here is my work around. It depends on dplyr, stringr and tidyr.

There are 2 limitations that are okay for my use, but might not for all (but you should be able to adapt the code if needed): 
1) It requires that all columns be read in as character.
[...]

PR #585, opened 2017-01-24
Add error for overlapping fwf specification (#534).

gergness:
I'm worried that the error message part of #534 got lost in the discussion of the full implementation for overlapping columns. The previous behavior can be very difficult to track down, and I think this error message helps tremendously.
Thanks for all the great work!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@
 
 * Pass `guess_max` to  fwf_empty, to control how many lines are read for whitespace before determining column structure.
 
+* `read_fwf()` gives error message if specifications have overlapping columns (#534, @gergness)
+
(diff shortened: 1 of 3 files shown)


hadley:
That's great - thanks!

@jimhester will do the review the details

jimhester:
I think this looks good as is, the only thing it needs is a note in NEWS explaining the fix and mentioning the issue number and your GitHub handle.

gergness:
Great, thanks!

jimhester:
I merged this manually at https://github.com/tidyverse/readr/commit/251337636a7a364a7c55f4bd61639478d68a72a9, Thanks!

PR #587, opened 2017-01-26
Return full string if no numbers passed to numeric parser

### LINKED ISSUE #548: problems() not pulling up the exact text of a parsing failure
The example:

```
library(tidyverse)
x <- tribble(
  ~unit, ~value,
  "a"  ,    "1",
  "b"  ,     "(NA)")
write_tsv(x, path = "test.tsv")
```
Now try and parse with the incorrect NA specification (in the real application it's not obvious exactly what represents NA).
```
new_x <- read_tsv("test.tsv",
                  na = "",
                  col_types = cols(
                      unit = col_character(),
                      value = col_number()
                    ))
file.remove("test.tsv")

problems(new_x)
# A tibble: 1 × 4
#     row   col expected actual
#   <int> <chr>    <chr>  <chr>
# 1     2 value a number       
```

I would have expected this to come up:
```
# A tibble: 1 × 4
#   row     col expected actual
# <int>   <chr>    <chr>  <chr>
# 1   2   value a number   (NA) 
```
Is this the expected behavior?

jimhester:
Fixes #548

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@
 * `write_lines()` now accepts a list of raw vectors (#542).
 * ListCallback S6 Class to provide a more flexible return type for callback functions (#568, @mmuurr)
 * parsing problems now include the filename (#581).
+* Numeric parser now returns the full string if it contains no numbers (#548).
 
 * parsing problems in `read_delim()` and `read_fwf()` when columns are skipped using col_types now report the correct column name (#573, @cb4ds)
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,6 +1,7 @@
 # readr 1.0.0.9000
 
 * parsing problems in `read_delim()` and `read_fwf()` when columns are skipped using col_types now report the correct column name (#573, @cb4ds)
+* Numeric parser now returns the full string if it contains no numbers (#548).
```
hadley: Needs a new line

PR #590, opened 2017-01-31
Better error handling for guess_max

### LINKED ISSUE #588: guess_max = Inf
Almost always when I read text files I want it to use all the rows to infer types. However, setting guess_max to Inf, NA, or -1 causes errors:

```
> orpha_genes = read_csv(orpha_genes_disk_file, guess_max = Inf)
Error in .Call("readr_guess_types_", PACKAGE = "readr", sourceSpec, tokenizerSpec,  : 
  negative length vectors are not allowed
In addition: Warning message:
In guess_types_(datasource, tokenizer, locale, n = n) :
  NAs introduced by coercion to integer range
> orpha_genes = read_csv(orpha_genes_disk_file, guess_max = NA)
Error in .Call("readr_guess_types_", PACKAGE = "readr", sourceSpec, tokenizerSpec,  : 
  negative length vectors are not allowed
> orpha_genes = read_csv(orpha_genes_disk_file, guess_max = -1)
Error in .Call("readr_guess_types_", PACKAGE = "readr", sourceSpec, tokenizerSpec,  : 
  negative length vectors are not allowed
> orpha_genes = read_csv(orpha_genes_disk_file, guess_max = NULL)
Error in guess_types_(datasource, tokenizer, locale, n = n) : 
  expecting a single value
```

Note that the error for NA is incorrect. NA is not a negative vector.

Instead I do something like this:

`orpha_genes = read_csv(orpha_genes_disk_file, guess_max = 999999)`

One has to be careful about how many 9's one adds, because if one adds too many one gets:

```
> read_csv("data/affy_diseases.csv", guess_max = 999999999999999999999999999)
Error in .Call("readr_guess_types_", PACKAGE = "readr", sourceSpec, tokenizerSpec,  : 
  negative length vectors are not allowed
In addition: Warning message:
In guess_types_(datasource, tokenizer, locale, n = n) :
  NAs introduced by coercion to integer range
```

I tried reducing the number of 9's until I stopped getting that error. At some point, my entire system crashed instead! Unfortunately, I don't know how many 9's I tried, so now I'm weary about trying again. I tried this one both my virtual Ubuntu 16.10 machine and on my Windows 8.1 host. Both systems crashed.

In general, this parameter seems to be bug-prone.

jimhester:
Fixes #588 

I added another example showing how to read a dataset as character and using `type_convert()` to infer types based on the full dataset as well.

### DIFF
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -218,7 +218,7 @@ col_concise <- function(x) {
 
 col_spec_standardise <- function(file, col_names = TRUE, col_types = NULL,
                                  guessed_types = NULL,
-                                 comment = "", skip = 0, n = 1000,
+                                 comment = "", skip = 0, guess_max = 1000,
                                  tokenizer = tokenizer_csv(),
                                  locale = default_locale(),
                                  drop_skipped_names = FALSE) {
@@ -353,7 +353,7 @@ col_spec_standardise <- function(file, col_names = TRUE, col_types = NULL,
   if (any(is_guess)) {
     if (is.null(guessed_types)) {
       ds <- datasource(file, skip = skip, comment = comment)
-      guessed_types <- guess_types(ds, tokenizer, locale, n = n)
+      guessed_types <- guess_types(ds, tokenizer, locale, guess_max = guess_max)
     }
 
     # Need to be careful here: there might be more guesses than types/names
@@ -364,8 +364,18 @@ col_spec_standardise <- function(file, col_names = TRUE, col_types = NULL,
   spec
 }
 
-guess_types <- function(datasource, tokenizer, locale, n = 1000) {
-  guess_types_(datasource, tokenizer, locale, n = n)
+guess_types <- function(datasource, tokenizer, locale, guess_max = 1000, max_limit = .Machine$integer.max %/% 100) {
+  if (is.na(guess_max) || guess_max < 0) {
+    stop("`guess_max` must be a positive integer", call. = FALSE)
+  }
+
+  if (guess_max > max_limit) {
+    warning("`guess_max` is a very large value, setting to `", max_limit,
+      "` to avoid exhausting memory", call. = FALSE)
+    guess_max <- max_limit
+  }
+
+  guess_types_(datasource, tokenizer, locale, n = guess_max)
 }
 
 guess_header <- function(datasource, tokenizer, locale = default_locale()) {
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -366,8 +366,9 @@ col_spec_standardise <- function(file, col_names = TRUE, col_types = NULL,
 
 
 check_guess_max <- function(guess_max, max_limit = .Machine$integer.max %/% 100) {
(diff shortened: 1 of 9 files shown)


R/col_types.R
```diff
@@ -364,8 +364,18 @@ col_spec_standardise <- function(file, col_names = TRUE, col_types = NULL,
   spec
 }
 
-guess_types <- function(datasource, tokenizer, locale, n = 1000) {
-  guess_types_(datasource, tokenizer, locale, n = n)
+guess_types <- function(datasource, tokenizer, locale, guess_max = 1000, max_limit = .Machine$integer.max %/% 100) {
+  if (is.na(guess_max) || guess_max < 0) {
```
hadley: Probably should check it's a numeric of length 1 first

And then you probably should pull out into `check_guess_max()`

R/type_convert.R
```diff
@@ -33,6 +33,11 @@
 #'
 #' df <- data.frame(x = c("NA", "10"), stringsAsFactors = FALSE)
 #' str(type_convert(df))
+#'
+#' # Type convert can be used to infer types from an entire dataset
+#' type_convert(
+#'   read_csv(readr_example("mtcars.csv"),
```
hadley: Indenting is mildly wack

jimhester:
Ok addressed comments at 5f848cc

R/col_types.R
```diff
@@ -364,8 +364,27 @@ col_spec_standardise <- function(file, col_names = TRUE, col_types = NULL,
   spec
 }
 
-guess_types <- function(datasource, tokenizer, locale, n = 1000) {
-  guess_types_(datasource, tokenizer, locale, n = n)
+
+check_guess_max <- function(guess_max, max_limit = .Machine$integer.max %/% 100) {
+  if (!(length(guess_max) == 1 && is.numeric(guess_max) &&
```
hadley: This logic scares me a little

jimhester: I agree it is a bit confusing, any suggestions on a clearer way to express this?

hadley: Two options:

```
stopifnot(is.numeric(guess_max), length(guess_max) == 1)
```

Or De Morgan's the whole thing, which I _think_ gives you

```
if (length(guess_max) != 1 || !is.numeric(guess_max) ||
    !is_integerish(guess_max)) || is.na(guess_max) || guess_max <= 0) {
```

jimhester: How about this, the assert definition is slightly complicated because we want to evaluate the `...` lazily.

```r
assert <- function(msg, ...) {
  dots <- eval(substitute(alist(...)))
  for (x in dots) {
    if (!isTRUE(eval(x, parent.frame()))) {
      stop(msg, call. = FALSE)
    }
  }
  invisible(TRUE)
}

assert("`guess_max` must be a positive integer",
  length(guess_max) == 1, is.numeric(guess_max), is_integerish(guess_max),
  !is.na(guess_max), guess_max > 0)
```

hadley: Hmmmmm, I think we can just stick to an `if` statement instead of introducing a new syntactic form 😉

PR #591, opened 2017-01-31
Use tibble::as.tibble to construct tibbles

### LINKED ISSUE #538: Use tibble::tibble() instead of C++ code
to create a tibble, because tibble is going to warn + modify column names in case of improper encoding.

jimhester:
Fixes #538

### DIFF
--- a/src/Reader.cpp
+++ b/src/Reader.cpp
@@ -44,7 +44,7 @@ void Reader::init(CharacterVector colNames) {
 }
 
 RObject Reader::readToDataFrame(int lines) {
-  int rows = read(lines);
+  read(lines);
 
   // Save individual columns into a data frame
   List out(outNames_.size());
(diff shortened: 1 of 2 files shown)


src/Reader.cpp
```diff
@@ -53,7 +53,7 @@ RObject Reader::readToDataFrame(int lines) {
     out[j++] = collectors_[*it]->vector();
   }
 
-  out.attr("class") = CharacterVector::create("tbl_df", "tbl", "data.frame");
+  out.attr("class") = CharacterVector::create("data.frame");
```
hadley: For consistency with haven, I think I'd prefer:

```cpp
    static Function as_tibble("as_tibble", Environment::namespace_env("tibble"));
    return as_tibble(out);
```

jimhester: 👍

PR #592, opened 2017-01-31
`parse_factor()` gains a `include_na` argument

### LINKED ISSUE #541: parse_factor should always generate explicit NAs
Or it should at least be an option

jimhester:
Which includes `NA` in the factor levels.

Fixes #541

### DIFF
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -200,6 +200,7 @@ guess_parser <- function(x, locale = default_locale()) {
 #'
 #' @param levels Character vector providing set of allowed levels.
 #' @param ordered Is it an ordered factor?
+#' @param include_na Should `NA` be included as a factor level?
 #' @inheritParams parse_atomic
 #' @inheritParams tokenizer_delim
 #' @inheritParams read_delim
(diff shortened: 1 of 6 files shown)


R/collectors.R
```diff
@@ -200,6 +200,7 @@ guess_parser <- function(x, locale = default_locale()) {
 #'
 #' @param levels Character vector providing set of allowed levels.
 #' @param ordered Is it an ordered factor?
+#' @param include_na Should `NA` be included as a factor level?
```
hadley: I think this should be, "If `NA` are present, include as an explicit factor to level?".  And I think it should default to `TRUE`

tests/testthat/test-parsing-factors.R
```diff
@@ -22,3 +22,7 @@ test_that("NAs silently passed along", {
   expect_equal(x, factor(c("a", "b", NA)))
 })
 
+test_that("NAs included in levels if desired", {
+  x <- parse_factor(c("a", "b", "NA"), levels = c("a", "b"), include_na = TRUE)
```
hadley: Can you also include a test with `levels = NULL`?

jimhester: I can once https://github.com/tidyverse/readr/pull/602 is done

R/collectors.R
jimhester: The current implementation includes NA regardless if it occurs in the input, I assume you would want it to only be included if the input contains NA?

hadley: Right

jimhester:
Closing in favor of https://github.com/tidyverse/readr/pull/617

PR #599, opened 2017-02-06
Use a datasource for all file arguments in read_table

### LINKED ISSUE #552: `read_table` can't read from pipes
The following does not work:

```
con <- pipe("perl -le 'print qq{x $_} for 1..5'")
read_table(con)
## Error in isOpen(con) : invalid connection
```

The reason seems to be that `read_table` first calls `fwf_empty`, which opens the connection and reads from it; then it calls `col_spec_standardise`, which also tries to open the same connection & read from it.  The latter fails because it's already been read from and can't be re-opened.

There's also another subsequent call to `datasource` (on line 47 in the current Git `master` of `read_table.R`) that would fail after this one.

In general, to support connections & pipes, each reading function should ensure that its file/connection is opened exactly once.  Avoiding these re-opens should give better performance too.

jimhester:
Fixes #552 

The `missing()` calls in https://github.com/tidyverse/readr/compare/552?expand=1#diff-7f0d4a03d8bda22664cf51f3fcf4a243R36 are a little gross, but we only want to override the current values if the first value is a data source _and_ non-default values are passed to the `datasource()` constructor.

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -33,6 +33,8 @@
 #' close(con)
 datasource <- function(file, skip = 0, comment = "") {
   if (inherits(file, "source")) {
+    if (!missing(skip)) { file$skip <- skip }
+    if (!missing(comment)) { file$comment <- comment }
     file
   } else if (is.connection(file)) {
     datasource_connection(file, skip, comment)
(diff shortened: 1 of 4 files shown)


R/source.R
```diff
@@ -33,6 +33,8 @@
 #' close(con)
 datasource <- function(file, skip = 0, comment = "") {
   if (inherits(file, "source")) {
+    if (!missing(skip)) { file$skip <- skip }
```
hadley: Can you put your comment here?  And use more new lines?

PR #602, opened 2017-02-07
`parse_factor()` now can accept `levels = NULL`

### LINKED ISSUE #497: col_factor() that automatically figures out levels
A nice feature to the `col_factor()` function in `col_types` would be to allow empty arguments. 
This should be treated similar to `data$variable <- factor(data$variable)` on character variables after using `read_csv`. That is, `col_factor()` would be `stringsAsFactors=TRUE` for the specific variables/columns.

jimhester:
This allows one to generate factor levels based on the data

Fixes #497

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 * parsing problems now include the filename (#581).
 * Numeric parser now returns the full string if it contains no numbers (#548).
 * `read_table()` can now handle `pipe()` connections (#552).
+* `parse_factor()` now can accept `levels = NULL`, which allows one to generate factor levels based on the data (like stringsAsFactors = TRUE) (#497).
 
 * parsing problems in `read_delim()` and `read_fwf()` when columns are skipped using col_types now report the correct column name (#573, @cb4ds)
 
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -198,7 +198,9 @@ guess_parser <- function(x, locale = default_locale()) {
 #' `parse_factor` is similar to [factor()], but will generate
 #' warnings if elements of `x` are not found in `levels`.
 #'
-#' @param levels Character vector providing set of allowed levels.
+#' @param levels Character vector providing set of allowed levels. if `NULL`,
+#'   will generate levels based on the unique values of `x`, ordered by order
+#'   of appearance in `x`.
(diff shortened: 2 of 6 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # readr 1.0.0.9000
 
+* `parse_time()` now can accept `levels = NULL`, which allows one to generate factor levels based on the data (like stringsAsFactors = TRUE) (#497).
```
hadley: I think you mean `parse_factor()` 😉

R/collectors.R
```diff
@@ -198,7 +198,8 @@ guess_parser <- function(x, locale = default_locale()) {
 #' `parse_factor` is similar to [factor()], but will generate
 #' warnings if elements of `x` are not found in `levels`.
 #'
-#' @param levels Character vector providing set of allowed levels.
+#' @param levels Character vector providing set of allowed levels. if `NULL`,
+#' will generate levels based on the unique values of `x`.
```
hadley: Need indent

R/collectors.R
```diff
@@ -216,6 +217,9 @@ guess_parser <- function(x, locale = default_locale()) {
 #'
 #' # parse_factor generates a warning & problems
 #' x2 <- parse_factor(x, levels)
+#'
+#' # Using an argument of `NULL` will generate levels based on values of `x`
+#' x2 <- parse_factor(x, levels = NULL)
```
hadley: How will the levels be ordered?

jimhester: I guess it is implementation dependant based on std::map, how should we order them, just sorted alphabetically?

jimhester: According to SO we can depend on it being alphabetically http://stackoverflow.com/questions/7648756/is-the-order-of-iterating-through-stdmap-known-and-guaranteed-by-the-standard

jimhester: And actually you are right, they are ordered by how they appear in the input, forgot how I implemented this.

jimhester:
Yes that is right, the levels will be ordered how they appear in the input.

jimhester:
Closing in favor of https://github.com/tidyverse/readr/pull/617

PR #608, opened 2017-02-09
read_table using any whitespace as a delimiter

### LINKED ISSUE #607: read.table compatibility
We need to have a function? (or argument to `read_delim()`) that lets us mimic the behaviour of base R, i.e. using any number of whitespace as a delimiter.

cc @kevinushey @javierluraschi

jimhester:
Fixes #607 

This implementation replaces the current implementation of `read_table()`. Not sure if we want to do that or preserve the old behavior as `read_table()` or another function.


hadley:
I think we need to preserve existing `read_table()`; I'd rather make this `read_delim(delim = "")` or some other special sentinel

jimhester:
could we call this `read_ws_delim()` or `read_table_ws()` or some variant of that? A sentinel value that uses a completely different tokenizer seems like surprising behavior to me.

hadley:
Oh true, the behaviour wrt quotes etc is totally different.  Maybe to echo `read_csv` maybe just `read_table2()`?

R/tokenizer.R
```diff
@@ -122,3 +122,12 @@ tokenizer_log <- function() {
 tokenizer_fwf <- function(begin, end, na = "NA", comment = "") {
   structure(list(begin = begin, end = end, na = na, comment = comment), class = "tokenizer_fwf")
 }
+
+#' @export
+#' @rdname Tokenizers
+#' @param begin,end Begin and end offsets for each file. These are C++
```
hadley: Included by mistake?

jimhester: Yep thanks

PR #609, opened 2017-02-09
Writing to compressed files / connections

### LINKED ISSUE #348: Writing compressed files
Is there currently any way to write to a compressed text file?

Ideally, compression would be automatically detected from the `path` (ending in `.gz`, `.bz2`, or `.xz`) argument of `readr::write_tsv()`.

jimhester:
Fixes #348 

This still needs documentation, more tests and examples, but works pretty well.

The connection writing uses [boost::iostreams](http://www.boost.org/doc/libs/1_63_0/libs/iostreams/doc/), so is nicely buffered and should be pretty performant.

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -102,7 +102,7 @@ read_connection <- function(con) {
   read_connection_(con)
 }
 
-standardise_path <- function(path, check = TRUE) {
+standardise_path <- function(path, input = TRUE) {
   if (!is.character(path))
     return(path)
 
@@ -122,7 +122,7 @@ standardise_path <- function(path, check = TRUE) {
     }
   }
 
-  if (isTRUE(check)) {
+  if (isTRUE(input)) {
     path <- check_path(path)
   }
   switch(tools::file_ext(path),
@@ -130,8 +130,13 @@ standardise_path <- function(path, check = TRUE) {
     bz2 = bzfile(path, ""),
     xz = xzfile(path, ""),
     zip = zipfile(path, ""),
-    path
-  )
+
+    # Use a file connection for output
+    if (!isTRUE(input)) {
+      file(path, "")
+    } else {
+      path
+    })
 }
 
 source_name <- function(x) {
--- a/R/source.R
+++ b/R/source.R
@@ -113,6 +113,7 @@ standardise_path <- function(path, input = TRUE) {
     if (requireNamespace("curl", quietly = TRUE)) {
       con <- curl::curl(path)
     } else {
+      message("`curl` package not installed, falling back to using `url()`")
       con <- url(path)
     }
     if (identical(tools::file_ext(path), "gz")) {
--- a/R/source.R
+++ b/R/source.R
@@ -110,10 +110,15 @@ standardise_path <- function(path) {
     return(path)
 
   if (is_url(path)) {
-    if (identical(tools::file_ext(path), "gz")) {
-      return(gzcon(curl::curl(path)))
+    if (requireNamespace("curl", quietly = TRUE)) {
+      con <- curl::curl(path)
     } else {
-      return(curl::curl(path))
+      con <- url(path)
+    }
+    if (identical(tools::file_ext(path), "gz")) {
+      return(gzcon(con))
+    } else {
+      return(con)
     }
   }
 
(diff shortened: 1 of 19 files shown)


tests/testthat/test-write-delim.R
```diff
@@ -92,3 +92,12 @@ test_that("does not writes a tailing .0 for whole number doubles", {
 
   expect_equal(format_tsv(tibble::data_frame(x = -123456789)), "x\n-123456789\n")
 })
+
+test_that("write_csv can write to compressed files", {
+  mt <- read_csv(readr_example("mtcars.csv.bz2"))
+
+  filename <- file.path(tempdir(), "mtcars.csv.bz2")
+  on.exit(unlink(filename))
+  write_csv(mt, filename)
```
hadley: What turns this path into a `bzfile()` connection?

jimhester: https://github.com/tidyverse/readr/pull/609/files#diff-fd81d97b84af37e70673ba40af498d75R144

hadley: I still don't get it 😦  Shouldn't there be a switch on the file extension that creates a connection given a path?

hadley: Oooooh it's standardise path. Duh

hadley:
It might be worth benchmarking to see the overhead of the connection interface. If it's not high, we could consider pouring everything through connections.

jimhester:
From my tests It seems like the overhead for connections is quite low, so I think we could safely handle all output through connections without trouble.

hadley:
In that case, I'd say go ahead and pour everything through the connection interface. That should become our standard approach.

src/write_connection.cpp
```diff
@@ -0,0 +1,41 @@
+#include "write_connection.h"
+
+#define class class_name
+#define private private_ptr
+#include <R_ext/Connections.h>
+#undef class
+#undef private
+
+#if R_CONNECTIONS_VERSION != 1
```
jimhester: It was added in https://github.com/wch/r-source/commit/41b64b12f36851cb87e9132612591d623c217443, and running `git branch --contains 41b64b12f36851cb87e91326125` on the r-source clone returns

```
  R-3-3-branch
  build
  build-trunk
  tags/R-3-0-3
  tags/R-3-1-3
  tags/R-3-2-3
  tags/R-3-2-4
  tags/R-3-2-4-revised
  tags/R-3-2-5
  tags/R-3-3-0
  tags/R-3-3-1
  tags/R-3-3-2
  travis
* trunk
```

So looks like the earliest release was 3.0.3

tests/testthat/test-write-delim.R
jimhester: Yeah that is a good point, I will think about a test to verify it is a bz2 file, probably the best way would be to check the magic number I think.

R/write.R
```diff
@@ -95,7 +93,7 @@ format_delim <- function(x, delim, na = "NA", append = FALSE, col_names = !appen
   stopifnot(is.data.frame(x))
 
   x <- lapply(x, output_column)
-  stream_delim(x, "", delim, col_names = col_names, append = append, na = na)
+  stream_delim(x, NULL, delim, col_names = col_names, append = append, na = na)
```
jimhester: Unfortunately not, `stdout()` is a text-mode connection and the connections API only works with binary mode connections, you get `'write' not enabled for this connection` errors if you try to use `stdout()` here.

hadley: Oh yeah :(

src/write_connection.cpp
hadley: Ok, awesome

R/source.R
```diff
@@ -102,29 +102,41 @@ read_connection <- function(con) {
...
+standardise_path <- function(path, input = TRUE) {
   if (!is.character(path))
     return(path)
 
   if (grepl("\n", path))
     return(path)
 
   if (is_url(path)) {
+    if (requireNamespace("curl", quietly = TRUE)) {
+      con <- curl::curl(path)
```
hadley: Bump

jimhester: I put the message below for the fallback, you are saying you want the message if we are using curl instead?

hadley: Sorry, I missed that because of the stupid issues threading. All looks good :)

PR #616, opened 2017-02-18
Add fwf_cols function

jrnold:
This adds a helper function `fwf_cols` that is a more intuitive way of specifying fixed width column start and end points. While `fwf_positions` requires three vectors for start, end, and names, `fwf_cols` accepts a named list of length-2 vectors of the column start and end positions.

For example, 
```r
# 3. Paired vectors of start and end positions
read_fwf(fwf_sample, fwf_positions(c(1, 30), c(10, 42), c("name", "ssn")))
# 4. Named list of start and end positions
read_fwf(fwf_sample, fwf_cols(list(name = c(1, 10), ssn = c(30, 42))))
```


hadley:
I think this is an improvement, but I wonder if a wrapper around tribble would be even nicer.

jrnold:
I was thinking about whether a wrapper around a data frame would be useful and almost included a version in it, but decided against it. My thinking was there's two main way you could get the column specifications (1) if the column specifications are a data frame with two (variable, widths) or three columns (variable, start, end), or (2) they are entering it by hand.

If the column specifications are already in a data frame (i'll call is `cols`), it's not too unclear to simply call the existing functions by simply referencing the columns of the data frame:
```r
fwf_postions(cols$start, cols$end, cols$varname)
``` 
To me, that's still pretty clear, and not too much typing.
Possibly `fwf_positions` and `fwf_widths` could be made into generic functions, with methods for the first function being a vector, matrix, or data frame.

The second case is entering it by hand (when it's not too many columns). In that case, having the columns as argument names and the widths or (start, end) as values seems most natural. `fwf_cols` could be generalized to allow the values to be either (start, end) tuples or widths.
```r
# with widths
fwf_cols(foo = 1, bar = 5)
# with (start, end) tuples
fwf_cols(foo = c(1, 4), bar = c(5, 10)
```

This came up when I was helping a student read a fixed-width file. I foolishly didn't RTFM before writing code, and assumed that the format was something like what I was just wrote. When we got an error and actually read the documentation, I was too lazy to adjust change the code and used `map` to convert it into what `fwf_positions` was looking for.

hadley:
What if we allowed `col_positions` to take a data frame? If one row, the values are widths; if two rows, the values are start and end. If >= 3 rows, an error. 

Then you'd have:

```R
read_fwf(fwf_sample, tibble(name = c(1, 10), ssn = c(30, 42)))
read_fwf(fwf_sample, tibble(name = 10, skip = 20, ssn = 12))
```

jrnold:
I don't know. It seems more natural and easy to document that `col_positions` accepts a "long" tibble like `fwf_positions` produces. And the helper functions, `fwf_positions`, `fwf_widths`, `fwf_cols` all return such a tibble. This would minimize the amount of code that's rewritten, and keep these functions backwards compatible while improving the ease of use.

If a user is able to write the following, it's about as concise as the code above, and I'd say as readable.
```r
read_fwf(fwf_sample, fwf_cols(name = c(1, 10), ssn = c(30, 42)))
read_fwf(fwf_sample, fwf_cols(name = 10, skip = 20, ssn = 12))
```
And the following would still work:
```r
x <- tribble(
  ~ col_name, ~start, ~ end
  name, 1, 10,
  ssn, 30, 42
)
read_fwf(fwf_sample, x)
```

jrnold:
Now I have it so that 

- The `col_positions` argument of`read_fwf` accepts a list with either `begin` and `end` columns or `width` and treats them appropriately
- The `fwf_cols(...)` calls `tibble(...)` or uses the first argument if it is a list. It will expect either 1 row, in which case it calls `fwf_widths` or 2 rows, in which case it calls `fwf_positions`.

R/read_fwf.R
```diff
@@ -20,23 +27,39 @@
...
-#' # You can specify column positions in three ways:
+#' # You can specify column positions in four ways:
 #' # 1. Guess based on position of empty columns
 #' read_fwf(fwf_sample, fwf_empty(fwf_sample, col_names = c("first", "last", "state", "ssn")))
 #' # 2. A vector of field widths
 #' read_fwf(fwf_sample, fwf_widths(c(20, 10, 12), c("name", "state", "ssn")))
 #' # 3. Paired vectors of start and end positions
 #' read_fwf(fwf_sample, fwf_positions(c(1, 30), c(10, 42), c("name", "ssn")))
+#' # 4. Named arguments with start and end positions
+#' read_fwf(fwf_sample, fwf_cols(name = c(1, 10), ssn = c(30, 42)))
```
hadley: Can you include the width form here too?

R/read_fwf.R
```diff
@@ -20,23 +27,39 @@
...
 read_fwf <- function(file, col_positions, col_types = NULL,
                      locale = default_locale(), na = c("", "NA"),
                      comment = "", skip = 0, n_max = Inf,
                      guess_max = min(n_max, 1000), progress = show_progress()) {
   ds <- datasource(file, skip = skip)
   if (inherits(ds, "source_file") && empty_file(file)) {
-    return(tibble::data_frame())
+    return(tibble::tibble())
+  }
+  if (!is.list(col_positions)) {
```
hadley: This feels too complicated to me. If we have `fwf_cols()` I don't think we need to worry about list/data.frame inputs.

R/read_fwf.R
```diff
@@ -20,23 +27,39 @@
...
+  } else if (c("width" %in% names(col_positions))) {
+    col_positions <- fwf_widths(col_positions$width,
+                                col_names = col_positions$col_names)
+  } else {
+    stop("col_positions must contain either 'begin' and 'end' columns",
+         " or a 'width' column.", stop. = FALSE)
   }
 
-  tokenizer <- tokenizer_fwf(col_positions$begin, col_positions$end, na = na, comment = comment)
+  tokenizer <- tokenizer_fwf(col_positions$begin, col_positions$end, na = na,
```
hadley: Can you change this back please?

R/read_fwf.R
```diff
@@ -1,3 +1,10 @@
+fwf_col_names <- function(nm, n) {
```
hadley: This should be much lower in the file

R/read_fwf.R
```diff
@@ -88,27 +107,55 @@ fwf_empty <- function(file, skip = 0, col_names = NULL, comment = "", n = 100L)
...
     begin = start - 1,
     end = end, # -1 to change to 0 offset, +1 to be exclusive,
     col_names = col_names
   )
 }
+
+
+#' @rdname read_fwf
+#' @export
+#' @param ... If the first element is a data frame,
```
hadley: This feels too flexible to me. But if you really think it's a good idea to keep it, the function signature should be `x, ...`

R/read_fwf.R
```diff
@@ -88,27 +107,55 @@ fwf_empty <- function(file, skip = 0, col_names = NULL, comment = "", n = 100L)
...
+    x <- x[[1]]
+  }
+  x <- try(lapply(x, as.integer), silent = TRUE)
+  if (inherits(x, "try-error")) {
+    stop("All elements in x or ... must be coercible to integer vectors.")
+  }
+  names(x) <- fwf_col_names(names(x), length(x))
+  x <- tibble::as_tibble(x)
+  if (nrow(x) == 2) {
+    fwf_positions(as.integer(x[1, ]),
```
hadley: Indenting style

R/read_fwf.R
```diff
@@ -88,27 +107,55 @@ fwf_empty <- function(file, skip = 0, col_names = NULL, comment = "", n = 100L)
...
+#'   positions.
+#'   Otherwise, the elements of `...` are used to construct a data frame
+#'   with or or two rows as above.
+fwf_cols <- function(...) {
+  x <- list(...)
+  # If first element is a list (including data frame), then
+  if (is.list(x[[1]])) {
+    x <- x[[1]]
+  }
+  x <- try(lapply(x, as.integer), silent = TRUE)
```
hadley: I don't think like approach. I'd say just let the error bubble up to the user.

tests/testthat/test-read-fwf.R
```diff
@@ -127,6 +127,43 @@ test_that("error on empty spec (#511, #519)", {
   expect_error(read_fwf(txt, pos), "Zero-length.*specifications not supported")
 })
 
+# fwf_cols
+test_that("fwf_cols produces correct fwf_positions object with elements of length 2", {
+  expected <- fwf_positions(c(1,  9, 4),
```
hadley: Can you please fix the indenting here too?

If the arguments don't fit on one line it should look like:

```
function_name(
   arg1,
   argument_name = arg2,
   ...
)
```

jrnold: So this is a different style than the one in http://adv-r.had.co.nz/Style.html, which would be
```r
function_name(arg1,
              argument_name = arg2,
              ...)
```

NEWS.md
```diff
@@ -1,5 +1,31 @@
 # readr 1.1.0
 
+* `fwf_cols()` allows for specifying the `col_positions` argument of
```
hadley: I think something went wrong with your merge 😞

jrnold: Oops. wtf did that merge do? Bad git :-( Sorry about that, and fixed now.

jimhester:
Thanks!

PR #617, opened 2017-02-21
Improvements to factor parsing

jimhester:
This includes the changes in https://github.com/tidyverse/readr/pull/602, https://github.com/tidyverse/readr/pull/592 as well as the fix for https://github.com/tidyverse/readr/issues/615.

### DIFF
--- a/tests/testthat/test-parsing-factors.R
+++ b/tests/testthat/test-parsing-factors.R
@@ -30,19 +30,20 @@ test_that("levels = NULL (497)", {
 })
 
 test_that("NAs included in levels if desired", {
-  x <- parse_factor(c("NA", "b", "a"), levels = c("a", "b", NA), include_na = TRUE)
-  expect_equal(x, factor(c(NA, "b", "a"), exclude = NULL))
+  x <- parse_factor(c("NA", "b", "a"), levels = c("a", "b", NA))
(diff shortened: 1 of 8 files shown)


tests/testthat/test-parsing-factors.R
```diff
@@ -17,8 +17,39 @@ test_that("warning if value not in levels", {
...
+
+  expect_equal(n_problems(x), 0)
+  expect_equal(x, factor(c("a", "b", "c", "b")))
+})
+
+test_that("NAs included in levels if desired", {
+  x <- parse_factor(c("NA", "b", "a"), levels = c("a", "b", NA), include_na = TRUE)
+  expect_equal(x, factor(c("NA", "b", "a"), exclude = NULL))
+
+  expect_warning(x <- parse_factor(c("NA", "b", "a"), levels = c("a", "b"), include_na = TRUE))
```
hadley: I don't think this should generate a warning.

jimhester: Why not? you are saying you want to include NA in the levels, but don't actually provide a NA level...

hadley: I think NA is a special case. But I might be overthinking it

jimhester: Ok thinking about this more I think you are right, `include_na = TRUE` now will include the NA level if there is an NA in the input and it is not in the levels.

PR #624, opened 2017-02-24
Reverse dependency results

jimhester:
Doesn't look to me like there are any failures related to readr changes.


hadley:
Might be worth double checking with gutenbergr?

jimhester:
Looks like it is failing on CRAN as well https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/gutenbergr-00check.html

PR #630, opened 2017-03-03
change message behavior of read_table

jrnold:
When using `read_table` I was surprised to get a print out of the `spec` even after specifying the `col_type` argument. This is different than `read_csv` and most of the other functions in `readr`.

Currently the behavior of `read_table` is to print the spec whenever `progress = TRUE`. This changes the behavior to be similar to that of the other functions in **readr**: print if `col_types` is `NULL`.


jimhester:
Looks good, please add a note to NEWS.md with the change and we can merge it.

jimhester:
Thanks!

PR #646, opened 2017-04-05
List files with readr_example() and remove internal keyword

jennybc:
I think it's nice to list the possible examples when `path` is unspecified. I also expected to see `readr_example()` in the help index. I think it's fine to remove the internal keyword, yes?


jimhester:
Thanks!

PR #651, opened 2017-04-18
Format source with clang-format and enforce it's use on travis

jimhester:


jimhester:
This is somewhat of a test to verify it is working correctly.

src/Collector.cpp
```diff
@@ -18,12 +18,14 @@ CollectorPtr Collector::create(List spec, LocaleInfo* pLocale) {
     return CollectorPtr(new CollectorDouble(pLocale->decimalMark_));
   }
   if (subclass == "collector_number")
-    return CollectorPtr(new CollectorNumeric(pLocale->decimalMark_, pLocale->groupingMark_));
+    return CollectorPtr(
```
hadley: Is there an option to put the closing parens on a new line?

src/Collector.cpp
```diff
@@ -58,16 +62,17 @@ std::vector<CollectorPtr> collectorsCreate(ListOf<List> specs,
 
 // Implementations ------------------------------------------------------------
 
-void CollectorCharacter::setValue(int i, const Token& t) {
-  switch(t.type()) {
+void CollectorCharacter::setValue(int i, const Token &t) {
```
hadley: I don't understand why it puts (what to me is) a type modifier in the object name

src/Collector.cpp
```diff
@@ -58,16 +62,17 @@ std::vector<CollectorPtr> collectorsCreate(ListOf<List> specs,
...
+  switch (t.type()) {
   case TOKEN_STRING: {
     boost::container::string buffer;
     SourceIterators string = t.getString(&buffer);
 
     if (t.hasNull())
       warn(t.row(), t.col(), "", "embedded null");
 
-    SET_STRING_ELT(column_, i, pEncoder_->makeSEXP(string.first, string.second, t.hasNull()));
+    SET_STRING_ELT(column_, i, pEncoder_->makeSEXP(string.first, string.second,
```
hadley: I'd prefer something more like

```cpp
SET_STRING_ELT(
  column_, i, 
  pEncoder_->makeSEXP(string.first, string.second, t.hasNull()));
```

Any chance that's an option?

hadley:
Brain dump of format options that I think I like:

* BracketAlignmentStyle: BAS_AlwaysBreak
* AllowShortBlocksOnASingleLine: false
* BinPackArguments: false
* BinPackParameters: false
* ConstructorInitializerAllOnOneLineOrOnePerLine: true

I couldn't see an option to make the closing parens always go on a new line.

src/Collector.cpp
jimhester: I don't see anything for that style, no

src/Collector.cpp
jimhester: I think relevant options are `DerivePointerAlignment: true` (which will use the most common alignment of `&` and `*`, or `PointerAlignment: Left` which would always align pointers / references with the left.

jimhester:
http://cf.monofraps.net/ Seems like a good site for playing with options interactively.

I changed the formatting with your suggested options, let me know if there are other options we might want to change.

hadley:
LGTM

Issue #306, opened 2015-11-03
Support multibyte encodings (e.g. UTF-16LE)


bilydr:
Hi, I am trying to read in a file with UTF-16LE encoding 

which can be done with base package codes  

``` R
df <- read.delim(file1, stringsAsFactors = FALSE, fileEncoding = 'UTF-16LE')
```

but when I try to use readr to do the same 

``` R
df <- read_tsv(file1, locale = locale(encoding = 'UTF-16LE'))
```

I got the error **Error: Incomplete multibyte sequence**

Can you please help fix it? Thanks for your advice!

hadley:
Can you please provide a reproducible example?

bilydr:
Hi, please see the data file and codes in folder below

https://www.dropbox.com/sh/6vw6irfe0m3f5bm/AACmdssmln4iZjxkpzbE1MEaa?dl=0

thanks!

Longyi Bi

bilongyi@gmail.com
Mobile: +33-668655292

bilydr:
and the screen shot
![screenshot](https://cloud.githubusercontent.com/assets/8634933/10937958/eef76f12-82f6-11e5-8eb6-db0c8986f547.PNG)

bilydr:
Hi any follow-up on this issue? thanks!

multiphrenic:
having a similar problem. even when I try loading the file using a file connection with explicit encoding, the read fails:
data.table(read_csv(file("data/Parking_Tags_data_2008.csv", encoding="UCS-2LE"))) for example, reads garbage, while read.csv works fine.

zeehio:
I simplified the example so it only needs R to work (no external files required).


    tmp_file_name <- tempfile()
    # This is the Byte Order Mark:
    rawbom <- raw(2)
    rawbom[1] <- as.raw(255)
    rawbom[2] <- as.raw(254)
    # This is the text. 
    text <- "1\t2\n"
    # Converted to UTF-16LE
    text_utf16 <- iconv(text,from="UTF-8",to="UTF-16LE", toRaw = TRUE)
    # Write the BOM and the text to a file
    fd <- file(tmp_file_name, "wb")
    writeBin(rawbom, fd)
    writeBin(text_utf16[[1]], fd)
    close(fd)
    df_test <- data.frame(V1=1, V2=2)
    works <- read.delim(tmp_file_name, header=FALSE, fileEncoding = "UTF-16LE")
    assertthat::are_equal(works[1,], df)

[...]

PR #673, opened 2017-05-11
Provide encoding information to the Source class

zeehio:
This PR starts dealing with #306.

Basically, if we want to fully support UTF-16 --and other encodings-- then the tokenizer in readr needs to be aware of the text encoding. So, for starters, the `source` class needs to provide the encoding to the tokenizer.

In this PR I just fix the `datasource` object (and the underlying C++ "Source" classes) so they provide the encoding to the tokenizer. It was going to be a much smaller PR, but the C++ Source class also parses comments found in the file (at the header), so this PR also needs to provide the `CodePointIterator` class I wrote, that can iterate through code points instead of raw bytes. 

Instead of: "1 byte == 1 character" with Unicode encodings there are code units, code points and graphemes. See http://stackoverflow.com/a/27331885/446149 for a simple and good description of these terms. While the most faithful approach to text parsing is to iterate through graphemes, here we stay at the approximation "1 grapheme == 1 code point". This approximation is good enough for the tokenizers in readr, as the graphemes we are interested in detecting (e.g. `\r`, `\n`, `,`, `;`...) work well under that approximation and other graphemes won't be affected.

As you will see if you review the PR, the detection of comments in the Source class is done in the encoding of the source, so the whole file does not need to be converted. I just need to convert the `comment` value to whatever the encoding is in the original file.

Additionally the source can now provide more than one comment pattern, so once the tokenizers are also adapted, we will be able to tell readr to skip all lines that start with either "#" or "//" easily.

The `CodePointIterator` class takes two `const char*` (`begin` and `end`) and an encoding. And it allows to iterate through the code points between `begin` and `end`. Improvements or changes to this class may have a significant impact to `readr` performance, especially once the tokenizers use it. For instance, we could make things faster by "checking less" and assuming files are properly encoded.

As a final note/rant, the `datasource_string` R function may have issues with the encoding of the string. The way `CHARSXP` represents the encoding (as "UTF-8", "latin1", "bytes", "native" or "unknown/any") is annoying, as it is hard for me to know what "native" may be, and it is confusing that "unknown" is the result when the string is pure "ASCII". Tips are welcome.

### DIFF
--- a/R/file.R
+++ b/R/file.R
@@ -27,8 +27,12 @@ read_file <- function(file, locale = default_locale()) {
   if (empty_file(file)) {
     return("")
   }
-
-  ds <- datasource(file, encoding = locale$encoding)
+  if (missing(locale)) {
+    encoding <- NULL
+  } else {
+    encoding <- locale$encoding
+  }
+  ds <- datasource(file, encoding = encoding)
   read_file_(ds, locale)
 }
 
@@ -39,7 +43,7 @@ read_file_raw <- function(file) {
     return(raw())
   }
 
-  ds <- datasource(file)
+  ds <- datasource(file, encoding = "bytes")
   read_file_raw_(ds)
 }
 
--- a/R/file.R
+++ b/R/file.R
@@ -28,7 +28,7 @@ read_file <- function(file, locale = default_locale()) {
     return("")
   }
 
-  ds <- datasource(file)
+  ds <- datasource(file, encoding = locale$encoding)
   read_file_(ds, locale)
 }
 
--- a/R/source.R
+++ b/R/source.R
@@ -53,15 +53,24 @@ datasource <- function(file, skip = 0, comment = "", encoding = default_locale()
 
     file
   } else if (is.connection(file)) {
+    if (missing(encoding)) {
+      encoding <- "bytes"
+    }
     datasource_connection(file, skip, comment, encoding)
   } else if (is.raw(file)) {
     datasource_raw(file, skip, comment, encoding)
   } else if (is.character(file)) {
-    if (grepl("\n", file)) {
+    if (is_literal_data(file)) {
+      if (missing(encoding)) {
+        encoding <- encoding_from_literal_data(file)
+      }
       datasource_string(file, skip, comment, encoding)
     } else {
       file <- standardise_path(file)
       if (is.connection(file)) {
+        if (missing(encoding)) {
+          encoding <- "bytes"
+        }
         datasource_connection(file, skip, comment, encoding)
       } else {
         datasource_file(file, skip, comment, encoding)
@@ -72,6 +81,17 @@ datasource <- function(file, skip = 0, comment = "", encoding = default_locale()
   }
 }
 
+is_literal_data <- function(text) {
+  is.character(text) && grepl("\n", text)
+}
+
+encoding_from_literal_data <- function(text) {
+  enc <- Encoding(text)
+  if (enc == "unknown") {
+    return("latin1") # assume text is ASCII, and latin1 is faster than UTF-8
+  }
+  enc
+}
 
 # Constructors -----------------------------------------------------------------
 
@@ -80,7 +100,8 @@ new_datasource <- function(type, x, skip, comment = "", encoding, ...) {
     class = c(paste0("source_", type), "source"))
 }
 
-datasource_string <- function(text, skip, comment, encoding) {
+datasource_string <- function(text, skip, comment,
+                              encoding = encoding_from_literal_data(text)) {
   new_datasource("string", text, skip = skip, comment = comment, encoding = encoding)
 }
 
@@ -90,10 +111,16 @@ datasource_file <- function(path, skip, comment, encoding) {
 }
 
 datasource_connection <- function(path, skip, comment, encoding) {
+  if (missing(encoding)) {
+    encoding <- "bytes"
+  }
   datasource_raw(read_connection(path), skip, comment = comment, encoding = encoding)
 }
 
 datasource_raw <- function(text, skip, comment, encoding) {
+  if (missing(encoding)) {
+    encoding <- "bytes"
+  }
   new_datasource("raw", text, skip = skip, comment = comment, encoding = encoding)
 }
 
--- a/R/source.R
+++ b/R/source.R
@@ -12,6 +12,9 @@
 #'    Literal data is most useful for examples and tests. It must contain at
 #'    least one new line to be recognised as data (instead of a path).
 #' @param skip Number of lines to skip before reading data.
+#' @param comment A string or character vector used to identify comments. Any
+#'                text after the comment characters will be silently ignored.
+#' @param encoding The text encoding of the data given in `file`. "UTF-8" as default
 #' @keywords internal
 #' @export
 #' @examples
--- a/R/source.R
+++ b/R/source.R
@@ -14,7 +14,10 @@
 #' @param skip Number of lines to skip before reading data.
 #' @param comment A string or character vector used to identify comments. Any
 #'                text after the comment characters will be silently ignored.
-#' @param encoding The text encoding of the data given in `file`. "UTF-8" as default
+#' @param encoding The text encoding of the data given in `file`. If `NULL` is
+#'                 passed (the default), then "UTF-8" is used for files,
+#'                 "bytes" is used for connections and raw vectors, and the
+#'                 string Encoding will be used for literal data strings.
 #' @keywords internal
 #' @export
 #' @examples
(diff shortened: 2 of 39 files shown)


jimhester:
`datasource_string()` is mainly used for testing, so I think it is safe to assume ASCII or possibly `UTF-8` if the encoding is not explicitly given.

zeehio:
I would rather work on integrating a grapheme iterator in the future. While the grapheme iterator is "the right way to do it" and that library looks good, I don't have time right now to work on it that and I don't see an urgent need for it. :sweat_smile: 

If you prefer that I squash all the commits feel free to say it. Right now all tests pass and I feel happy with the current status of the patch.

jimhester:
Don't worry about squashing, we will do it when we merge. Hadley will likely want to review this and may have some additional comments. You will need to add a note to NEWS.md detailing the change and referencing the issue # and your GitHub handle.

zeehio:
I have added the NEWS bullet as suggested. I have also refactored the CodePointIterator class so the encodings are more independent and adding support for another encoding is easier (if needed).

R/file.R
```diff
@@ -27,8 +27,12 @@ read_file <- function(file, locale = default_locale()) {
   if (empty_file(file)) {
     return("")
   }
-
-  ds <- datasource(file)
+  if (missing(locale)) {
```
hadley: This code is repeated in a lot of places. Could you pull out into a separate function?  

I'm also not sure why you have different logic when the locale is missing.

R/source.R
```diff
@@ -12,6 +12,12 @@
 #'    Literal data is most useful for examples and tests. It must contain at
 #'    least one new line to be recognised as data (instead of a path).
 #' @param skip Number of lines to skip before reading data.
+#' @param comment A string or character vector used to identify comments. Any
+#'                text after the comment characters will be silently ignored.
+#' @param encoding The text encoding of the data given in `file`. If `NULL` is
```
hadley: I don't understand why these should be different

src/Source.cpp
```diff
@@ -10,16 +11,26 @@ SourcePtr Source::create(List spec) {
   std::string subclass(as<CharacterVector>(spec.attr("class"))[0]);
 
   int skip = as<int>(spec["skip"]);
-  std::string comment = as<std::string>(spec["comment"]);
+  SEXP comment_sexp = spec["comment"];
```
hadley: Why not `CharacterVector comment_sexp`? That will throw an error if not correct type, and simplifies the for loop that follows.

src/Source.h
```diff
@@ -9,47 +11,93 @@ typedef boost::shared_ptr<Source> SourcePtr;
...
+  inline std::string encoding() {
+    if (pEncoding == "") {
+      stop("Encoding not set at Source");
+    } else {
+      return pEncoding;
+    }
+  }
+  inline const std::vector<std::string>& get_comments() { return pComments; }
+
+  void set_comments(std::vector<std::string> utf8comments) {
```
hadley: Why do we need these methods?

src/Source.h
```diff
@@ -103,10 +171,20 @@ class Source {
   static SourcePtr create(Rcpp::List spec);
 
 private:
-  static bool
-  inComment(const char* cur, const char* end, const std::string& comment) {
-    boost::iterator_range<const char*> haystack(cur, end);
-    return boost::starts_with(haystack, comment);
+  std::string pEncoding;
```
hadley: I don't think our style prefixes private variables. @jimhester can you please confirm?

R/source.R
zeehio: At first I thought that having "UTF-8" as default was the best approach. This is what `default_locale()$encoding` gives, so it makes sense to keep it everywhere as default.

However, when working with datasources that are literal strings (like `"à b\n1 2\n3 4"`) I found it made more sense to (by default) autodetect the encoding using `Encoding()` if possible. This forced me to change the default value from `"UTF-8"` to `NULL` and autodetect it when a literal string is passed.

Then there was `datasource_connection`. If for instance the connection is a pipe, the default encoding usually is the one given by the locale. Here I assumed "bytes", meaning "something ASCII compatible that I will not convert". If I had chosen "UTF-8" then errors like "invalid UTF-8 character" would appear when someone used a pipe giving non-UTF8 characters, breaking backwards compatibility in those cases.

For `datasource_raw` I used a similar reasoning: "bytes" does not check for UTF-8 validity and most encodings are ASCII compatible, so it will preserve better backwards compatibility.

Having a default value for each type of datasource maybe complicates things too much, as it then forced me to check other functions calling `datasource()` and set the new default `NULL` encoding if it was not given.

If you prefer I can simplify this, keeping "UTF-8" as the only default in all cases.

src/Source.h
zeehio: The `set_comments` is used by the SourceString.h, SourceRaw.h and SourceFile.h classes. It takes care of storing the comments and, if an encoding was given, convert them to that encoding. I can replace the `set_comments` call with an assignment and `reencode_comments()` and remove that method if you want.

The Source class had the `begin()` and `end()` virtual methods that return a `const char *begin_` or `end_` in all derivative classes. Having something like `comments()` (the `get_` part maybe is unnecessary) or `encoding()` makes sense to me for consistence.

src/Source.h
zeehio: I saw some `pWarnings` and `pLocale`... I also saw `comment_`, `NA_` and `pEncoder_`. Having **p**ublic, **p**rotected and **p**rivate makes a prefix with `p` quite useless anyway.

zeehio:
I agree with you that this PR is too big and maybe too spread. If it is too much to take at once feel free to ask me to close this PR and start again (with more PR and smaller changes) so there are less questions to answer on each PR. It may take longer to me to prepare the PR but maybe the reviews and merges are faster.

Once this part is merged I am planning to submit a PR for the encoding support in the tokenizers. I will split the tokenizers in smaller PR so the review is easier.

src/Source.h
jimhester: pWarnings and pLocale were signifying they were pointers to those objects. The convention used for private variables is the`_` suffix.

hadley:
Yeah, I think it would be useful to break this up into a few places:

* Implementing `CodePointIterator` and then using in the appropriate places
* Making comment parsing the responsibility of the source, and supporting multiple prefixes
* Guessing the initial encoding from the source

(not necessarily in that order). That'll make it easier for us to review.

I'd also appreciate some benchmarks to assess how much of an impact this has on performance.

zeehio:
Thanks @hadley and @jimhester for your feedback. I will close this PR and submit these smaller pieces.

Regarding the benchmarking, the Tokenizer method that does most of the parsing is `nextToken()` so I expect an impact once the `CodePointIterator` is used on the tokenizers, but not because of changes in Source class.  In `Tokenizer::nextToken()`, we will move from "const char* dereferences and comparisons" to "function calls to methods that do comparisons and basic arithmetic". The only way I can think of to avoid this performance drop is using C++ templates, so the compiler generates optimized code for each Tokenizer & encoding. In `src/TokenizerDelim.cpp` we would have something like:

```
template <enum Encoding>
class TokenizerDelim: public Tokenizer {
...
Token nextToken() {
 ....
 CodePointIterator<Encoding> cur(...);
 ....
}
```

Where `Encoding` is an `enum` for the supported `CodePointIterator`s. I believe that with that approach we could solve all the performance issues at compile time, at least for the most common and basic "ASCII encoding" case.

src/Source.cpp
zeehio: Feel free to postpone this discussion if you want, but if we solve this now then I will do it right in the PR.

I don't see the benefit of using a `CharacterVector` here, as I need to set the encoding to UTF-8 anyway. This is a simple example (for a single element) using SEXP and CharacterVector. I will use the CharacterVector anyway if you prefer.

```
library(Rcpp)
cppFunction('std::string test1(SEXP x) {
  SEXP x_elm = STRING_ELT(x, 0);
  std::string x_elm_conv = Rf_translateCharUTF8(x_elm);
  Rcpp::Rcout << x_elm_conv << std::endl;
  return x_elm_conv;
}')

cppFunction('std::string test2(CharacterVector x) {
  Rcpp::String x_elm(x(0));
  x_elm.set_encoding(CE_UTF8);
  Rcout << x_elm.get_cstring() << std::endl;
  return x_elm.get_cstring();
}')


x <- "potato\u00f3"
test1(x)
test2(x)


x2 <- "potato\xf3"
Encoding(x2) <- "latin1"
test1(x2)
test2(x2)
```

hadley: Well you never actually check that `comment_sexp` is a character vector and I'm not sure `STRING_ELT(comment_sexp, i)` checks its inputs. So you at least need a test for typeof.

PR #676, opened 2017-05-13
Don't test error message that is created by tibble

krlmlr:
As discussed in https://github.com/krlmlr/readr/commit/4d1021e7ab09a66bd432b25590603d7c7f6baf88#commitcomment-22129539. I've stripped the CRAN comments because they don't apply anymore, and the version number update. Thanks again for your help.


jimhester:
Thanks!

PR #677, opened 2017-05-16
Make the Source class report the encoding

zeehio:
This pull request is the first of a series to provide multibyte encoding support to readr on the tokenizers. I made a first attempt in #673, but that PR was too big to be easily reviewed.

With this pull request, the datasource() R function and the associated C++ Source class now will require an encoding. If this encoding is ambiguous in endianness (like UTF-16 or UTF-32, which mandate a Byte Order Mark) the BOM is detected and skipped (as happened before) and the encoding is updated to reflect the endianness (UTF-16LE, UTF16-BE...).

Future pull requests will use this encoding for proper source parsing.

As this PR is not a new feature (not yet at least) and it is not a bug fix, I have not added a NEWS entry. Feel free to ask for it if you feel it is better to have it.

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -5,6 +5,14 @@ collectorGuess <- function(input, locale_) {
     .Call(`_readr_collectorGuess`, input, locale_)
 }
 
+source_encoding <- function(spec) {
+    .Call(`_readr_source_encoding`, spec)
+}
+
+whitespaceColumns <- function(sourceSpec, n = 100L, comment = "") {
+    .Call(`_readr_whitespaceColumns`, sourceSpec, n, comment)
+}
+
 read_connection_ <- function(con, chunk_size = 64 * 1024L) {
     .Call(`_readr_read_connection_`, con, chunk_size)
 }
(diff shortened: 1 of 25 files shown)


zeehio:
Once this is merged I will submit another PR, that offers the possibility of multiple comments and doing the comment parsing in the Source class (instead of duplicated code in Source and Tokenizers):

- https://github.com/zeehio/readr/tree/multiple_comment_prefixes

jimhester:
First, I wanted to thank you for working on this PR and the previous ones. This was not a simple problem and I appreciate the effort you went through to solve it.

I also wanted to apologize for leaving this open for so long, it was certainly not my intent to keep you waiting, but other projects and priorities kept me busy.

We have plans to begin moving away from the legacy parser in readr, in favor of the parsers in [vroom](https://vroom.r-lib.org/) or possibly [arrow](https://github.com/apache/arrow/tree/master/r). Because of this I don't think it would be prudent to add this amount of new code to the existing parser.

Even though we're closing this issue, we really appreciate your work, and hope you'll continue to contribute in the future.

Issue #671, opened 2017-05-10
write_csv: scientific notation cannot be disabled


dpprdan:
`write_csv()` turns (some) longer numbers into scientific notation and there does not seem to be a way to disable it. This has been [mentioned before](https://github.com/tidyverse/readr/issues/229#issuecomment-138295522) in #229 and apparently was fixed then, so this might be a regression?

``` r
library("readr")
df <- data.frame(a = -0.0004029971, b = 0.0412975501857025)
print(df, digits = 17)
#>                         a                    b
#> 1 -0.00040299710000000002 0.041297550185702497
cat(format_csv(df))
#> a,b
#> -4.029971e-4,0.0412975501857025
```

The problem is I cannot use scientific notation (well) with the tools that import the csv.

Also compare [this](https://github.com/tidyverse/readr/issues/229#issuecomment-138296578) to this (which should be equivalent IMHO):

``` r
format_csv(data.frame(GEOID = seq(from = 60150001022000, to = 60150001022005, 
  1)))
#> [1] "GEOID\n60150001022e3\n60150001022001\n60150001022002\n60150001022003\n60150001022004\n60150001022005\n"
```

I.e. `GEOID\n60150001022e3` instead of `GEOID\n60150001022000`

jimhester:
I don't think we will be changing this behavior in the near future (if ever). A workaround you can use is to format the columns before writing. See `?base::format` for details on possible formatting arguments.

``` r
format_numeric <- function(x, ...) {
  numeric_cols <- vapply(x, is.numeric, logical(1))
  x[numeric_cols] <- lapply(x[numeric_cols], format, ...)
  x
}

library("readr")
df <- data.frame(a = -0.0004029971, b = 0.0412975501857025)
format_csv(format_numeric(df))
#> [1] "a,b\n-0.0004029971,0.04129755\n"
```

dpprdan:
Thanks! One general question though: Why default to a notation/formatting that, at least to me, seems to be less compatible with other tools? (Even more so when the file format, csv, is arguably one of the most interchangeble/compatible formats.) I guess this is all a matter of perspective, I just would like to understand your design choice.

One addition and one question with respect to the format\_numeric fuction: I guess one ought to add the `scientific = FALSE` option to reliably disable scientific notation, irrespective of `options(scipen)`.

``` r
format_numeric_jh <- function(x, ...) {
  numeric_cols <- vapply(x, is.numeric, logical(1))
[...]

Issue #229, opened 2015-08-04
Loss of precision with csv_write


donboyd5:
Hi. I hope I'm doing this correctly. I see that this is the same as #117, which was closed, but I think the issue remains. There is a reproducible example below:

```
adf <- data.frame(a=0.04129755018570249)
print(adf, digits = 17)
write.csv(adf, "e:/temp/adf_base.csv")
write_csv(adf, "e:/temp/adf_readr.csv")

print(read_csv("e:/temp/adf_base.csv")$a, digits=17)
print(read_csv("e:/temp/adf_readr.csv")$a, digits=17)
```

hadley:
Yeah, I know about that, but I forgot to re-open.  It's surprisingly difficult to recreate what R does (particularly eliminating non-sigificant FP differences)

laurikoobas:
There is loss of precision the other way as well - longer numbers are turned into scientific notation for write_csv() and I haven't found a way to disable it. Nobody else has either : http://stackoverflow.com/questions/30341140/readr-turn-off-scientific-notation-in-write-csv

hadley:
@laurikoobas Looks fine to me:

```R
 write_csv(data.frame(GEOID = seq(from=60150001022000, to=60150001022005, 1)), "")
#> [1] "GEOID\n60150001022000\n60150001022001\n60150001022002\n60150001022003\n60150001022004\n60150001022005\n"
```

laurikoobas:
It does indeed. The difference is between CRAN version and dev version. I guess it got fixed after 0.1.1 release.

hadley:
@donboyd5 actually I think this is ok now:

``` r
tmp <- tempfile()

adf <- data.frame(a=0.04129755018570249)
print(adf, digits = 17)
#>                     a
#> 1 0.04129755018570249
write.csv(adf, tmp)
print(read_csv(tmp)$a, digits=17)
#> [1] 0.041297550185702497

write_csv(adf, tmp)
print(read_csv(tmp)$a, digits=17)
#> [1] 0.041297550185702497
```

donboyd5:
This is great, thank you. Yes, I have been testing it and everything seems good. It was all that was holding me up from using readr. Thanks for yet another great productivity tool. Don

dpprdan:
@hadley: [This](https://github.com/tidyverse/readr/issues/229#issuecomment-138295522) is not completely fixed, I think.

```r
> df <- data.frame(a=-0.0004029971, b=0.04129755018570249)
[...]

PR #679, opened 2017-05-20
Allow control of scientific notation in write_csv

zeehio:
Issues #671 and #229  show that there are users that want to control how scientific notation is used when writing CSV files.

Currently readr (using the grisu3 C library) decides that integers ending with more than two zeroes are printed using scientific notation [in this line of code](https://github.com/tidyverse/readr/blob/master/src/grisu3.c#L364). This usually makes sense from a readability point of view, but gives non-uniform results in cases like this:

```
cat(format_csv(data.frame(a = seq(from=1998, to=2002, 1))))
a
1998
1999
2e3
2001
2002
```

For cases like this, readability is better if scientific notation is avoided.
While using integers instead of doubles already prints numbers correctly, some
users use doubles to store integers that can be larger than 2^32 (see the linked issues).

This pull request offers the possibility of disabling scientific notation on
positive integers. If this option is enabled, integer numbers up to 10^15 will
be printed without scientific notation. The 10^15 limit is reasonable, as
2^53+1 (approx 9*10^15) is the first integer that can't be represented as double without losing
accuracy.

```
cat(format_csv(data.frame(a = seq(from=1998, to=2002, 1)), int_use_scientific = FALSE))
a
1998
1999
2000
2001
2002
```

Another loosely related issue is the fact that `write_csv` does not offer a
`scipen`-like argument, that some users want because other CSV import tools don't
cope well with scientific notation 
[link to question (stackoverflow)](https://stackoverflow.com/questions/30341140/readr-turn-off-scientific-notation-in-write-csv).

This PR also addresses this issue, by offering a `scipen` argument. Currently the tradeoff
between using fixed point notation or scientific notation is based on a tradeoff
between the length of the number and its readability. With positive values of scipen
we bias the decision to fixed point format (penalizing scientific notation), while
with a negative value we bias towards scientific notation. There are no absolutes,
as with this PR we don't allow to fully disable scientific notation. If we did we would print
very very long (>300 digits) numbers slow to parse.

```
df <- data.frame(a = c(1E-1, 1E-2, 1E-3, 1E-4))
format_csv(df, scipen = -3)
#[1] "a\n1e-1\n1e-2\n1e-3\n1e-4\n"
format_csv(df, scipen = -2)
#[1] "a\n0.1\n1e-2\n1e-3\n1e-4\n"
format_csv(df, scipen = -1)
#[1] "a\n0.1\n0.01\n1e-3\n1e-4\n"
format_csv(df, scipen = -0)
#[1] "a\n0.1\n0.01\n0.001\n1e-4\n"
format_csv(df, scipen = 1)
#[1] "a\n0.1\n0.01\n0.001\n0.0001\n"
```

Feel free to review and merge if you feel these features are worth it.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -23,3 +23,4 @@
 ^travis/check_format\.sh$
 ^src/\.clang-format$
 ^.vimrc$
+^CONDUCT\.md$
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,0 +1,25 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, we pledge to respect all people who 
+contribute through reporting issues, posting feature requests, updating documentation,
+submitting pull requests or patches, and other activities.
+
+We are committed to making participation in this project a harassment-free experience for
+everyone, regardless of level of experience, gender, gender identity and expression,
+sexual orientation, disability, personal appearance, body size, race, ethnicity, age, or religion.
+
+Examples of unacceptable behavior by participants include the use of sexual language or
+imagery, derogatory comments or personal attacks, trolling, public or private harassment,
+insults, or other unprofessional conduct.
+
+Project maintainers have the right and responsibility to remove, edit, or reject comments,
+commits, code, wiki edits, issues, and other contributions that are not aligned to this 
+Code of Conduct. Project maintainers who do not follow the Code of Conduct may be removed 
+from the project team.
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by 
+opening an issue or contacting one or more of the project maintainers.
+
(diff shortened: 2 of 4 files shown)


ThoDuyNguyen:
Currently i use hack from this post to bypass the issue 

https://stackoverflow.com/questions/30341140/readr-turn-off-scientific-notation-in-write-csv

Still hope this pull request will be accepted.

jimhester:
Thank you for the pull request. I am not sure we want to go down the route of having a number of customization options for writing, part of the idea is to use standardized formatting.

I think we should simply turn off scientific formatting of integers entirely.

zeehio:
Thanks for the review, I don't know what I did but I closed the pull request.

Anyway, I will resubmit a pull request to simply turn off scientific formatting of integers.

Issue #683, opened 2017-06-04
AccumulateChunkCallback


blakeboswell:
I recently had to build a linear regression model on ~50GB of data with the ability to hold about 12GB in memory at a time.  I was able to get the job done using `read_delim_chunked` and `biglm` as follows:

``` r
# fit_equation is a function that returns an equation built from data colnames
chunk_regress <- function(fit_equation){
  function(data, index, mdl){
    if(is.null(mdl))
      return(biglm(fit_equation(colnames(data)), data))
    update(mdl, data)
  }
}


AccumulateChunkCallback <- R6::R6Class("AccumulateChunkCallback",
     inherit = readr::ChunkCallback,
     private = list(
       acc = NULL
     ),
     public = list(
       initialize = function(callback) {
         private$callback <- callback
       },
       receive = function(data, index) {
         private$acc <- private$callback(data, index, private$acc)
       },
       result = function() {
         private$acc
       }
     )
)


mdl <- read_delim_chunked(
    file = file_path,
    callback = AccumulateChunkCallback$new(chunk_regress(fit_equation)),
    chunk_size = 2 * 10^6,
    delim = ',',
    col_types = list(.default = col_number())
)
```

Is there a smarter way to accomplish this pattern using one of the built-in call-backs?

I could see this pattern being useful more broadly when you need access to the result of the callback on earlier chunks while handling later chunks... without enough memory to rbind all the results.

PR #689, opened 2017-06-17
Add AccumulateCallback

blakeboswell:
This pr addresses issue #683. Judging by the issue's activity, the user community isn't clamoring for this capability. However, I encounter the need regularly enough to think others will find it useful.

## Summary

`AccumulateCallback` accumulates the results of operations on chunks in a single variable rather than collecting them in DataFrame or List.  This allows for a `reduce` operation to be performed without creating an intermediate object that may be too large to hold in memory.

### Example

In this example, we sum the `mpg` feature of `mtcars`.

``` R
f <- function(x, pos, acc) sum(x$mpg) + acc

total_mpg <- read_csv_chunked(readr_example("mtcars.csv"),
                              AccumulateCallback$new(f, acc = 0),
                              chunk_size = 5)
```

- The `callback` function requires an additional input, `acc` which enables operation results to be accumulated.
- The `initialize` method of `AccumulateCallback` also takes an additional optional input, `acc`, which allows for the initial value of the accumulator to be defined.  This value defaults to `NULL` if omitted.

### SideEffectChunkCallback Work Around

This same behavior can be accomplished with the `SideEffectChunkCallback` as follows

``` R
acc <- 0
f <- function(x, pos){
  acc <<- sum(x$mpg) + acc
}
read_csv_chunked(readr_example("mtcars.csv"),
                 SideEffectChunkCallback$new(f),
                 chunk_size = 5)
acc
```
With this in mind, this pr does not really provide a new capability.  Just a more explicit expression of an existing capability.

### Practical Example: Regression

A practical example would be fitting a glm to a data set that is too large to hold in memory.

``` R
library(biglm)

chunk_regress <- function(data, index, mdl){
  if(is.null(mdl))
    return(biglm(mpg ~ wt, data))
  update(mdl, data)
}

mdl <- read_csv_chunked(readr_example("mtcars.csv"),
                        AccumulateCallback$new(chunk_regress),
                        chunk_size = 5)

summary(mdl)
```

In this example we rely on the fact that the default `acc` value of `AccumulateCallback` `initialize` method is `NULL` to set the initial condition of `chunk_regress`.


jimhester:
This looks good, thanks!

Thanks again for writing the PR and for your patience in getting it merged!

PR #692, opened 2017-06-25
Add support for overlapping fixed width column specs

gergness:
Adds support for overlapping columns in C++ (as opposed to an R loop with multiple reads in #634). This isn't the most common format, but has popped up a few times (and is supported in other stats packages like Stata).

I'm not sure if the changes in RcppExports.R are spurious, but I got them when using both the current CRAN version of Rcpp and the development version.

Thank you!


jimhester:
Thank you for the PR and for your patience getting it merged!

PR #698, opened 2017-07-20
Correct documentation for type_convert about guessing

gergness:
Documentation assumes that the guessing in `type_convert()` relies on the first 1,000 rows as elsewhere. However, this does not appear to be true:

```r
library(readr)

test <- tibble::tibble(x = c(1:10000, "a"))
class(type_convert(test)$x)
#> Parsed with column specification:
#> cols(
#>   x = col_character()
#> )
#> [1] "character"
````


jimhester:
Thanks!

PR #699, opened 2017-07-24
Avoid warning in type_convert when providing non-character specs

gergness:
Before this commit, if I use a col_spec with non-character variables in type_convert, I get a warning.

```r
library(readr)
test <- tibble::data_frame(x = 1:3, y = "3", z = "a")
type_convert(test, col_types = cols(x = col_number()))

#> # A tibble: 3 x 3
#>       x     y     z
#>   <int> <dbl> <chr>
#> 1     1     3     a
#> 2     2     3     a
#> 3     3     3     a
#> Warning message:
#> The following named parsers don't match the column names: x 
```

### DIFF
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -365,6 +365,11 @@ col_spec_standardise <- function(file, col_names = TRUE, col_types = NULL,
 }
 
 
+col_spec_limit_names <- function(col_spec, names_only) {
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks for the PR!

I ended up fixing this issue somewhat differently, but this situation will no longer produce a warning.

PR #701, opened 2017-08-04
add fwf_header

lbartnik:
Hi @hadley and others,

I had to read an `fwf` file with a header line and I came up with this utility to compute file description based on where known column names match in the header line. Let's say you have a header line with the following columns (first line):

    a b c d e f
    1 2 3 4 5 6

and you want to read columns `a`, `b` and `e`. Call:

    read_fwf(file, skip = 1,
             col_positions = fwf_header(path, col_names = c("a", "b", "c", "e", "f")),
             col_types = cols_only(a = "i", b = "i", e = "i"))

You have to ask `fwf_header` to compute positions of columns `c` and `f` to match the end position of `b` and `e`. Other than that, I think it's pretty straightforward.

If you'd like to make this a part of `readr` I'll be happy to clean-up the code and add unit tests.

### DIFF
--- a/R/read_fwf.R
+++ b/R/read_fwf.R
@@ -14,12 +14,13 @@
 #' @inheritParams tokenizer_fwf
 #' @inheritParams read_delim
 #' @param col_positions Column positions, as created by [fwf_empty()],
-#'   [fwf_widths()] or [fwf_positions()]. To read in only selected fields,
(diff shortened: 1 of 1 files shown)


jimhester:
If I understand correctly I think `read_table()` is essentially what this is trying to do, it parses a fixed width file based on the whitespace in the header.

lbartnik:
This is a really old PR, I'm not sure if it's still relevant. I'm just going to close it.

PR #712, opened 2017-09-12
Correct reference to date_names_langs

cderv:
`date_names_locales` has been replace by `date_names_langs` for some time now. 

I believe it is the last reference to it in the documentation.


jimhester:
Thanks!

Issue #729, opened 2017-10-29
readr::read_csv issue: Chinese Character becomes messy codes


kk22boy:
I'm trying to import a dataset to RStudio, however I am stuck with Chinese characters, as they become messy codes. Here is the code:

```
library(tidyverse)
df <- read_csv("中文,英文\n英文,德文")
df
# A tibble: 1 x 2
  `\xd6\xd0\xce\xc4`            `Ӣ\xce\xc4`
               <chr>                  <chr>
1 "<U+04E2>\xce\xc4" "<U+00B5>\xc2\xce\xc4"
```

When I use the base function read.csv, it works well. I guess I must do something wrong with encoding. But there are no encoding option in read_csv, how can I do this?

yutannihilation:
I answered on SO: https://stackoverflow.com/a/46999569/5397672

BTW, I wonder why we have to explicitly specify the encoding here. `file` argument should be translated into UTF-8 before passing it to C++.

Sciurus365:
Chinese characters seem to be encoded using "GB2312" \:
`> df <- parse_character("中文,英文\n英文,德文")
> df
[1] "\xd6\xd0\xce\xc4,Ӣ\xce\xc4\nӢ\xce\xc4,�\xc2\xce\xc4"
> df <- parse_character("中文,英文\n英文,德文",locale=locale(encoding="GB2312"))
> df
[1] "中文,英文\n英文,德文"`
And it seems that the functions of readr::read_csv can not identify characters encoded by GB2312 automatically, while base::read,csv can.

jimhester:
readr assumes character input is UTF-8 unless you specify a different encoding explicitly. This is done on purpose to ensure code is reproducible across systems.

@Sciurus365 Thank you for tracking down what the encoding of the characters actually was, I think this can be closed.

yutannihilation:
@Sciurus365 Thanks for the information! But, I'm afraid this is not correct:

> while base::read.csv can.

Both `read.csv()` and `read_csv()` don't **identify** the encoding from the characters at all. Instead, they just **assume** the encoding (`read.csv()` bets on the system's default encoding and `read_csv()` bets on UTF-8) and, in this case, `read.csv()` won the bet.

However, while assuming and praying can be a good strategy for external resources, for character vectors that already on R's memory, I think we can **identify** their encoding by using the `Encoding()` attribute. This is the basic idea of my PR #730.

yutannihilation:
@jimhester 

[...]

PR #730, opened 2017-10-29
encode characters as UTF-8 when locale is not set

yutannihilation:
As reported on #729, `read_csv()` etc fails to handle the encoding on non-UTF-8 MBCS locale when `file` is a text:

```r
df <- read_csv("中文,英文\n英文,德文")
```

In this case, `"中文,英文\n英文,德文"` is encoded as the default encoding but the `default_locale()` assumes that `file` is encoded as UTF-8. So, it would be better to encode the text as UTF-8 before passing to other functions.

This PR fixes the problem, but I'm not 100% sure this PR is OK since this may break some existing use cases where `file` is a UTF-8 text marked improper encoding (`locale(encoding = "UTF-8")` is identical to `default_encoding()`):

```r
read_csv(utf8_but_improperly_encoded_text, locale = locale(encoding = "UTF-8"))
```

Users will have to encode the text properly before passing it to `read_csv()` etc.

```r
Encoding(properly_encoded_text) <- "UTF-8"
read_csv(properly_encoded_text)
```

What do you think?


R/read_delim.R
```diff
@@ -177,7 +177,13 @@ read_delimited <- function(file, tokenizer, col_names = TRUE, col_types = NULL,
     if (empty_file(file)) {
        return(tibble::data_frame())
     }
-    data <- file
+    if (is.character(file) && identical(locale, default_locale())) {
```
jimhester: I think this is a _rare_ case where `missing(locale)` might be the right thing to do.

yutannihilation: Oh, I didn't know `missing()` propagates to the function called inside... Thanks for the suggestion!

yutannihilation: Ah, no. In this case, `locale` never be "missing" because there are the default value for `locale` in the function that call `read_delimited()` (e.g. `read_csv()`, `read_table()`...).

``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2017-11-03

# this is ok
f1 <- function(x) missing(x)
f2 <- function(x) f1(x = x)
f2()
#> [1] TRUE

# missing() returns FALSE if x has the default value in the callee function
f1 <- function(x) missing(x)
f2 <- function(x = 1L) f1(x = x)
f2()
#> [1] FALSE
```

yutannihilation:
Although this PR may breaks some use cases, now I'm sure this is justifiable since

1. copying and pasting is the most primitive form of reproducibility, so `df <- read_csv("中文,英文\n英文,德文")` should return the same result regardless of which platform we execute the code on.
2. readr expects UTF-8 inputs, so it's the user's responsibility to either 
      1. encode the input into UTF-8 before passing it to `read_*()`. This PR is OK for this.
      2. specify the proper `encoding` for the input. This PR is *almost* OK for this; it fails if the input is actually UTF-8 but is marked some improper encoding. I bet this use case is rare and the problem can be avoidable by `Encoding<-()` or `iconv()`.

jimhester:
Ok I think this is fine, can you add a note to NEWS.md with the issue # and mentioning your GitHub username and I will merge it.

yutannihilation:
Thanks! I've added it.

jimhester:
Thanks!

PR #732, opened 2017-10-31
Support reading from the system clipboard using a file named "clipboard"

### LINKED ISSUE #656: Read from clipboard
I was wondering if reading from a clipboard will be supported in the future? There was an [earlier issue](https://github.com/tidyverse/readr/issues/112) which provided a work-around, but it doesn't seem to work anymore.

```
> x <- readr::read_csv(file("clipboard"))
Error in read_connection_(con) : can only read from a binary connection
> x <- readr::read_csv(file("clipboard", open = 'rb'))
Error in file("clipboard", open = "rb") : 
  'mode' for the clipboard must be 'r' or 'w'
```

read_csv can only read from a binary connection while the connection itself only allows for non-binary connections.

Using the base functions, this is quite straightforward.

```
> x <- read.csv("clipboard", sep = '\t')
> x
  a b c
1 1 2 3
2 4 5 6
```

jimhester:
This gives readr the same interface as base functions. Although here it is actually nicer because we use `clipr::read_clip()` which actually works on non-X11 based systems.

This PR also contains code to treat input of character vectors of greater than 1 as data to read. I included it mainly a thing I have missed having, and it deals with the same code I was modifying for the clipboard stuff.

Fixes #656


jimhester:
@hadley Do you have any reservations about reading from the clipboard with 'clipboard' or of accepting character vectors of length > 1 as lines of text to be read?

hadley:
I'd prefer to use a sentinel value like `clipboard()` I think.

Accepting character vectors seems like a good idea.

jimhester:
Agreed, `clipboard()` is nicer, and this way you can still read a file named 'clipboard'. Updated accordingly.

PR #744, opened 2017-11-14
Clarify `fwf_cols()` docs

dapperjapper:
I was having trouble with this sentence structure..


jimhester:
Yes I think that reads better as well. Thanks!

PR #758, opened 2017-12-06
[Bug Fix] `quoted_na` was not being used by `read_csv`

RomeroBarata:
Fix a bug in `read_csv` where the argument `quoted_na` was not being used.


jimhester:
Thanks!

PR #763, opened 2017-12-13
Add support for day of week in parse_datetime

tigertoes:
To keep with other similar libraries, I've chosen `%a` as the designator. This might cause some issues when doing non-abbreviated day of week in future as `%A` is already reserved for other uses.

This PR updates the documentation, however none of the generated docs have been updated - is this something expected of contributors? If so, could you please tell me how to generate the documentation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,6 +32,7 @@ you would pass `skip = 3`, now you only need to pass `skip = 2`.
 * `read_*()` files now support reading from the clipboard by using `clipboard()` (#656).
 * `write_file()` gains a `sep` argument, to specify the line separator (#665).
 * Allow files to be read via FTP over SSH by recognising `sftp` as a URL protocol (#707, @jdeboer).
+* `parse_date*() accepts `%a` for local day of week (#763, @tigertoes).
 
 ## Bug Fixes
 
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -257,7 +257,8 @@ col_factor <- function(levels, ordered = FALSE, include_na = FALSE) {
 #'     70-99 -> 1970-1999.
 #'   \item Month: "\%m" (2 digits), "\%b" (abbreviated name in current
 #'     locale), "\%B" (full name in current locale).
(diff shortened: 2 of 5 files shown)


jimhester:
Thanks, this package uses [roxygen2](https://cran.r-project.org/package=roxygen2) for documentation, so rather than editing the `.Rd` file you need to edit the relevant [code comment](https://github.com/tidyverse/readr/blob/master/R/collectors.R#L259), then run `devtools::document()` to update the `Rd` file.

If you don't have everything setup to run `devtools::document()` locally just change the comment and I can run it after I merge.

tigertoes:
Thanks Jim, I've generated the docs as you've instructed, however quite a lot of other files have changes as well. Could you look at my most recent commit (b74509be7) and make sure nothing has been put in that shouldn't please?

jimhester:
Thanks! I think the added changes are due to using the devel version of roxygen2 rather than the CRAN version. If you could `install.packages("roxygen2")` it should restrict the changes to only the single file.

If you could also add a test for when using `%a` with a date that fails to parse that would be helpful!

And please add a note to NEWS.md describing this change and mentioning your GitHub username and this PR #.

Thanks again for working on this!

tigertoes:
I'll do all those things, but may I suggest having a CONTRIBUTING file listing all those things out so future contributors know what is expected of them?

jimhester:
Yes, I agree and apologize for not having one already. We are working on a standardized one we can use in all tidyverse packages.

tigertoes:
Hi Jim, I've made the various changes you've asked for - are you able to take a look at this again please?

tigertoes:
@jimhester Hi mate, it's been nearly half a year since this PR had been looked at. I believe I've addressed all the issues raised, and it's now committed against HEAD as of time of this comment. Could you please take a look again and merge if you're happy?

tigertoes:
Closing this as I've heard no response for over half a year now.

jimhester:
Thanks for your patience and for submitting the pull request!

PR #766, opened 2017-12-17
Support multiple comment symbols

zeehio:
This pull request allows to provide more than one comment pattern to `readr`. Example:

```
readr::read_delim(c(
  "# A comment",
  "x,y",
  "1,2",
  "// another comment",
  "3,4"), delim = ",", comment = c("#", "//"))
```

It is applied on top of this other PR https://github.com/tidyverse/readr/pull/677, so merging this will be easier once that PR is reviewed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,12 @@
 * Allow files to be read via FTP over SSH by recognising `sftp` as a URL protocol (#707, @jdeboer).
 * `read_*()` now converts string `file`s to UTF-8 before parsing, which is convenient for non-UTF-8 platforms
   in most cases (#730, @yutannihilation).
+
+* Comments are now handled by the `datasource` instead of by the `tokenizer` (@zeehio, #766)
+
+* Allow a character vector of comments. Using `comment = c("//", "#")` will
+  skip all the lines starting with either `//` or `#`. (@zeehio, #766)
+
 # readr 1.1.1
 
 * Point release for test compatibility with tibble v1.3.1.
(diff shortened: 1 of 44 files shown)


jimhester:
I am reposting my comment from https://github.com/tidyverse/readr/pull/677 here, again I sincerely apologize but I don't see another option and I don't think it is fair to keep these PRs open when there is no real chance of them being merged.

First, I wanted to thank you for working on this PR and the previous ones. This was not a simple problem and I appreciate the effort you went through to solve it.

I also wanted to apologize for leaving this open for so long, it was certainly not my intent to keep you waiting, but other projects and priorities kept me busy.

We have plans to begin moving away from the legacy parser in readr, in favor of the parsers in [vroom](https://vroom.r-lib.org/) or possibly [arrow](https://github.com/apache/arrow/tree/master/r). Because of this I don't think it would be prudent to add this amount of new code to the existing parser.

Even though we're closing this issue, we really appreciate your work, and hope you'll continue to contribute in the future.

zeehio:
Hi @jimhester, I totally understand the situation. For me these PR were part of a learning process and it was a wonderful experience to interact with you and the rest of the team. It was worth it, even if in the end it never got merged. I understand the advantages of vroom (using ALTREP, etc) and I'm looking forward to keep contributing in the future as I've done in the past.

Besides, RStudio has my CV in case you ever decide to eventually hire more people and find me worth considering (it's a long shot but hey, it's always worth trying 😅)

Thanks for your time and nice words!

PR #777, opened 2018-01-16
Fix missing # in example causing check to fail

mawds:
A missing # in the col_types example was causing "them like so:" to be executed as code and so the package checks to fail.


jimhester:
Thanks!

PR #808, opened 2018-03-06
Tweak language for na parameter

kwstat:
Current wording says "Character vector of strings to use for missing values."  To me, this sounds like these will be INSERTED in place of missing values, so I changed the text to "to interpret as missing values". 

For what it's worth, read.csv uses 'na.strings', which is documented as "a character vector of strings which are to be interpreted as NA values".

Also, would it be worthwhile to caution the user that "NA" strings (North America, Sodium, etc) will be turned into missing values?


jimhester:
Thanks! I agree that is less ambiguous wording.

PR #838, opened 2018-05-01
convert path to the native locale before passing to boost::interprocess::file_mapping()

### LINKED ISSUE #834: `read_csv` can't recognize chinese file path on R 3.5.0
> `R.version.string`
[1] "R version 3.4.3 (2017-11-30)"
> `repaid <- read_csv("D:/work51/催收专员绩效m1/还款4月0425.csv", `
                   `col_types = list(advance_clear_amount = col_double()))`

> `R.version.string`
[1] "R version 3.5.0 (2018-04-23)"
> `repaid <- read_csv("D:/work51/催收专员绩效m1/还款4月0425.csv", `
                   `col_types = list(advance_clear_amount = col_double()))`
Error in guess_header_(datasource, tokenizer, locale) : Cannot read file D:/work51/鍌敹涓撳憳缁╂晥m1/杩樻4鏈?425.csv: 系统找不到指定的路径。

### LINKED ISSUE #837: `read_delim` can't recognize portuguese BR file path on R 3.5.0
```
DCV_Corporativo <- read_delim("Validação HD/DCV/Dados/DCV_Corporativo.txt", 
"|",
escape_double = FALSE,
col_types = cols(Data = col_datetime(format = "%d/%m/%Y")), 
trim_ws = TRUE,
skip = 2)
```

**output:**
Error in guess_header_(datasource, tokenizer, locale) : 
Cannot read file ValidaÃ§Ã£o HD/DCV/Dados/DCV_Corporativo.txt:
O sistema não pode encontrar o caminho especificado.

yutannihilation:
fixes #834, fixes #837 

As described on #834, the behavior of `normalizePath()` (more precisely, `path.expand()`, which is used inside `normalizePath()`) has been changed to return the character encoded in UTF-8. I guess this happens only with Windows on non-UTF-8 locales.

R 3.4.4 on Windows 10:

``` r
Encoding(normalizePath("~/鬼"))
#> [1] "unknown"
```

R 3.5.0 on Windows 10:

``` r
Encoding(normalizePath("~/鬼"))
#> [1] "UTF-8"
```

This usually doesn't become a problem; on R codes, most R functions take care of the encoding automatically. But, on C++ codes, the encoding attribute is lost when the `String` is converted to `std::string`. So, we need to convert it to the proper encoding before passing it to the functions that take `std::string`, in this case, `boost::interprocess::file_mapping()` here:

https://github.com/tidyverse/readr/blob/6f0bb65296afa55709fd60cdc5d59a4c89623e36/src/SourceFile.h#L19-L20

It seems `boost::interprocess::file_mapping()` takes the path string encoded in the **native locale** (`CE_NATIVE`). So, we need to ensure the string is encoded properly, instead of blindly pass the one passed from R session.

IIUC, Rcpp doesn't have a function to do this, so `Rf_translateChar()` is the choice in this case. (I'm not an Rcpp expert, so please point out if I'm wrong at this point....)

### DIFF
--- a/src/Source.cpp
+++ b/src/Source.cpp
@@ -18,8 +18,8 @@ SourcePtr Source::create(List spec) {
     return SourcePtr(
         new SourceString(as<CharacterVector>(spec[0]), skip, comment));
   } else if (subclass == "source_file") {
-    SEXP path(as<CharacterVector>(spec[0])[0]);
-    return SourcePtr(new SourceFile(Rf_translateChar(path), skip, comment));
+    CharacterVector path(spec[0]);
+    return SourcePtr(new SourceFile(Rf_translateChar(path[0]), skip, comment));
   }
 
   Rcpp::stop("Unknown source type");
--- a/src/Source.cpp
+++ b/src/Source.cpp
@@ -18,8 +18,8 @@ SourcePtr Source::create(List spec) {
     return SourcePtr(
         new SourceString(as<CharacterVector>(spec[0]), skip, comment));
   } else if (subclass == "source_file") {
-    std::string path(as<CharacterVector>(spec[0])[0]);
-    return SourcePtr(new SourceFile(path, skip, comment));
+    SEXP path(as<CharacterVector>(spec[0])[0]);
+    return SourcePtr(new SourceFile(Rf_translateChar(path), skip, comment));
   }
 
   Rcpp::stop("Unknown source type");
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,6 +51,7 @@ you would pass `skip = 3`, now you only need to pass `skip = 2`.
   in most cases (#730, @yutannihilation).
 * `read_*()` no longer throws a "length of NULL cannot be changed" warning when
   trying to resize a skipped column (#750, #833).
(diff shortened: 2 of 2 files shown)


yutannihilation:
@jimhester Do you also prefer `enc2native(normalizePath())` solution here as with the case of readxl?

src/Source.cpp
```diff
@@ -18,8 +18,8 @@ SourcePtr Source::create(List spec) {
     return SourcePtr(
         new SourceString(as<CharacterVector>(spec[0]), skip, comment));
   } else if (subclass == "source_file") {
-    std::string path(as<CharacterVector>(spec[0])[0]);
-    return SourcePtr(new SourceFile(path, skip, comment));
+    SEXP path(as<CharacterVector>(spec[0])[0]);
```
jimhester: I think it is more clear if this is a CharacterVector rather than a SEXP.

yutannihilation:
Thanks for the comment about avoiding vague `SEXP`. I've made a change. Is this what you mean?

yutannihilation:
@jimhester Do I need to make more changes? If no, I think this is ready to merge now :)

jimhester:
Thanks!

yutannihilation:
Thanks for merging!

Issue #788, opened 2018-02-08
Consider using read_private instead of read_only for SourceFile


jimmo:
I don't have much background with R but I noticed this while trying to get some R code running in our sandbox which (for better or worse) uses a FUSE filesystem.

`read_delim()` was failing with `ENODEV`, which I tracked down to `SourceFile` attempting to mmap the file with `MAP_SHARED`. I'm fairly sure this would fail on a network filesystem too, as well as FUSE?

Is there any reason this needs to be a shared mapping? Could you use `MAP_PRIVATE` (via `boost::interprocess::read_private`) instead?

e.g.
https://github.com/tidyverse/readr/compare/master...groklearning:mmap-private?expand=1

Thanks

PR #841, opened 2018-05-04
Use private mapping (instead of shared) for SourceFile.

jimhester:
i.e. boost::interprocess::read_private instead of
boost::interprocess::read_only

Filesystems are much likely to support MAP_PRIVATE rather than
MAP_SHARED (e.g. FUSE).


jimhester:
Need to look into this further, but opened a PR to capture the change.

#788

PR #850, opened 2018-05-11
Link cheatsheet directly to raw PDF

batpigandme:
As discussed in https://github.com/tidyverse/readr/pull/846#issuecomment-388237538.


jimhester:
LGTM!

PR #867, opened 2018-06-18
Create CODEOWNERS

batpigandme:


jimhester:
I wonder if @jennybc should be the understudy for readr rather than @hadley, thoughts?

batpigandme:
Had that same thought – I was just following the sheet, but happy to change it up!

hadley:
Yeah, good idea!

jimhester:
Ok thanks!

PR #877, opened 2018-08-02
Use levels = NULL as default in parse_factor() (#862)

### LINKED ISSUE #810: parse_factor() is the only parser that lacks a single character shortcode
Now that `parse_factor()` can operate without explicit levels (#497), it seems like it should have a single letter shortcode. It is the only parser that does not.

### LINKED ISSUE #862: parse_factor() make levels = NULL the default
Since #497 `parse_factor()` has the option to automatically figure out levels by setting `levels = NULL`. Shouldn't this be the default, so that, for example, a call to `col_factor()` without arguments would make use of this feature?

I came across this because I was intuitively expecting `col_factor()` to do this already. Seems it wouldn't break any existing code since currently `col_factor()` alone throws an error.

mikmart:
Updated `parse_factor()` and `col_factor()` to use `levels = NULL` as default. Fixes #862.

This also fixes #810 by adding `"f"` as a shortcode for `col_factor()`, enabled by the default values.

### DIFF
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -241,6 +241,7 @@ col_concise <- function(x) {
     c = col_character(),
     D = col_date(),
     d = col_double(),
+    f = col_factor(),
     i = col_integer(),
     l = col_logical(),
     n = col_number(),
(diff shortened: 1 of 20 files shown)


jimhester:
Thanks, please add a note to NEWS.md and I can merge this!

PR #893, opened 2018-09-21
Use https for badge links.

QuLogic:
This would fix a mixed-content warning on https://readr.tidyverse.org/ However, note that there is still an old link to http://tidyverse.org/rstudio-logo.svg; I did not change those as they are in pkgdown-generated files. I assume that would be taken care of when the files are rebuilt.

(PS, can you also change the URL here on GitHub to https as well?)


jimhester:
Thanks!

PR #897, opened 2018-10-13
Expand on mention of compressed formats

### LINKED ISSUE #849: Better communicate the ability to write to compressed formats
In `read_csv()` we show an example of writing compressed output files, but never explain it in the help text, so this ability seems underused.

namelessjon:
This fixes #849 

Please let me know if you have any feedback


jimhester:
Thanks!

PR #902, opened 2018-10-21
Add function to read data from Google Sheets

jiho:
The changes induce some notes about the length of the URLs in the examples but I don't see how to fix those.

The examples are based on a Google Sheet I made. It would be better to transfer ownership to the maintainer of `readr` and ensure it will always be present and accessible.


jimhester:
Thank you for the pull request! However the [googlesheets](https://cran.r-project.org/web/packages/googlesheets/index.html) package serves this purpose and has existed on CRAN for a number of years, I do not feel adding (limited) functiontaily to readr would make sense at this time.

jiho:
As far as I can see, the googlesheets package (that I know and use) is really meant to deal with Google Sheet in *your google account*. To read an external Google Sheet with googlesheets, you needs to register it through its url, then *log in* to your google account (so you need to have one), and finally read its content through the sheets API (which, as far as my tests show, outputs the data in a form that is slightly different from what the File > Export to csv function does -- namely all fields, including numeric ones, are quoted in the API export for some reason while none is quoted in the .csv export).

The aim of this function is to allow to easily read the data from any Google Sheet into R (by just copy-pasting the url), allowing to iterate on editing the data in the sheet, reading it, plotting it, etc. and finally do the "right thing" and export the data in an open, future proof format (.csv, .tsv) and store it together with the code for the final analysis (which would then use `read_csv` rather than `read_gs`).
The code uses the export url rather than the API which means the data will look exactly the same as the final one, exported from Google Sheets.

So I still think there is a use case for this, which is separate from what googlesheets proposes, and it would have its place next to `read_csv` (which actually does the reading) into `readr`. Let me know what you think.

jimhester:
I think in that case we could just possibly include an example, in the `read_csv()` / `read_tsv()` documentation.

It seems all you need to do is add `export?format=tsv`, `export?format=csv` to the base URL and include that directly to the read function.

e.g.

``` r
readr::read_tsv("https://docs.google.com/spreadsheets/d/1bkLTSMLobkBjIPY0HaLJjW2J_shER3kTXP6SoXdER0k/export?format=tsv")
#> Parsed with column specification:
#> cols(
#>   model = col_character(),
#>   mpg = col_double(),
#>   cyl = col_double(),
#>   disp = col_double(),
#>   hp = col_double(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_double(),
#>   am = col_double(),
#>   gear = col_double(),
#>   carb = col_double()
#> )
#> # A tibble: 32 x 12
#>    model   mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>    <chr> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#>  1 Mazd…  21       6  160    110  3.9   2.62  16.5     0     1     4     4
#>  2 Mazd…  21       6  160    110  3.9   2.88  17.0     0     1     4     4
#>  3 Dats…  22.8     4  108     93  3.85  2.32  18.6     1     1     4     1
#>  4 Horn…  21.4     6  258    110  3.08  3.22  19.4     1     0     3     1
#>  5 Horn…  18.7     8  360    175  3.15  3.44  17.0     0     0     3     2
#>  6 Vali…  18.1     6  225    105  2.76  3.46  20.2     1     0     3     1
#>  7 Dust…  14.3     8  360    245  3.21  3.57  15.8     0     0     3     4
#>  8 Merc…  24.4     4  147.    62  3.69  3.19  20       1     0     4     2
#>  9 Merc…  22.8     4  141.    95  3.92  3.15  22.9     1     0     4     2
#> 10 Merc…  19.2     6  168.   123  3.92  3.44  18.3     1     0     4     4
#> # ... with 22 more rows
```

<sup>Created on 2018-10-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

jennybc:
For the record, you can use googlesheets and, especially, googlesheets4 in a non-authenticated way. googlesheets itself uses the csv link for data access whenever possible (i.e. user hasn't requested a specific sheet or cells).

If you add an example for this to readr, you'll probably want to detail what needs to be true about the Sheet in terms of permissions in order to read directly from the URL.

jiho:
OK @jennybc I stand corrected, sorry for the misunderstanding. I just happened that every time I tried (and I tried again just before posting) the googlesheets functions redirected me to Google's authentication page. So I assumed it was the required behaviour.

@jimhester indeed, it's just a wrapper for `read_csv()`. But, as you can see in the code of the pull-request, there are a couple subtleties to access a specific sheet within the spreadsheet (which, through the csv export requires the sheet id code and not the sheet name) or a specific range (but that's less of a concern because it will probably be used more scarcely).

Actually, my goal was to make this easy and ~ fool-proof for the user: you're sent a link to a gsheet (meaning that you should have at least View access to it, which is sufficient), you copy-paste the link as the function argument and it "just works".
I coded this function recently for students. They were working with data shared by all and gsheet is just unbeatable for this, and they were discovering R, through the tidyverse, so it just made sense to code it this way. I figured others could benefit from it and that's all ;-)

jiho:
If it's not worth making it into a function, then, indeed, an example of the basic case (reading the first sheet of a gsheet) using Jenny's Gapminder gsheet would already help. Let me know and I'll propose something.
(For myself I don't care; for students I would prefer the easy function-based solution ;-) )

jennybc:
This **is** a good thing to document. It just sits right at the intersection of readr and googlesheets(4) which makes it hard to place. Regardless, I should make this workflow nice in googlesheets4.

PR #913, opened 2018-11-19
Read connections to a temporary file rather than to a raw vector

jimhester:
Because we previously read connections into an in-memory vector we would
frequently run into out of memory issues when reading from a
connection.

We automatically cleanup the temporary file using a finalizer when the datasource object is removed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,6 +53,7 @@ implementation of the `melt_*()` functions!
 
 ## New features
 
+* Functions now read connections to a temporary file rather than to an in-memory object (#610, #76).
 * `melt_*()` functions added for reading ragged data (#760, @nacnudus).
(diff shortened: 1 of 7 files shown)


jimhester:
@hadley I realize this is a departure from how we have currently handled connections. It will likely be slower for cases when the data does fit in memory, but it at least makes it possible to handle out of memory data from connections in conjunction with the chunked readers.

It would be good to get your input if this the right thing to do, I feel like it is.

jimhester:
Also ideally we would support in-memory streaming but it would require more significant changes in how data is read from data sources, e.g. we would need something like an iterator interface rather than always assuming a contiguous byte array.

jimhester:
Thanks for reviewing!

PR #927, opened 2018-11-28
Updated documentation of col_types to include factors

osorensen:
Noted that `factor` was not part of the documentation to the `col_types` argument of `read_delim`, despite the fact that setting `col_types = "f"` yields a factor column. Hence suggesting an update of the documentation.


jimhester:
Thanks!

Issue #926, opened 2018-11-26
Addition of a `guess_delim` to guess the delimiter of a csv file


cimentadaj:
I have recently developed a prototype version of a `guess_delim` function that tries to guess the delimiter of a csv file. This work was inspired from [Sniffer](https://docs.python.org/2/library/csv.html#csv.Sniffer) from the csv module in Python, which tries to guess the delimiter of a csv file.

The function has been tested in a couple of thousand files in a package I'm developing for an API and it seems to have promising potential. I think this function would fit perfectly within the `readr` ecosystem as it would complement `guess_encoding`.

However, I think the biggest gain from this function would be that it would allow users to build custom functions that detect delimiters and dispatch accordingly to different `readr_*` functions. This pontential fits particularly well within `readr`'s philosophy because all `readr_*` functions have the same arguments, so a dispatch based on delimiter would work perfectly for any delimiter in terms of argument naming.

There is the option of including a function such as the one described above on `readr` but of course that would be more experimental from `readr`'s side.

If you think this would fit `readr` then I could create a branch and start adding this function while slowly refactoring the code to make it fit the coding style of `readr`. Alternatively, we could exchange in this issue and refactor directly in this thread and when we agree on a final function then I could move everything into a branch.

jimhester:
Could you add your function and make `guess_delim()` the default value for the `delim` parameter in `readr::read_delim()`?

e.g. `read_delim <- function(file, delim = guess_delim(file)`

In terms of dicussing the code please open a [Pull Request](https://help.github.com/articles/about-pull-requests/) and we can discuss it there.

jimhester:
Also in terms of heuristics for the guessing you might also want to look at how this is done in [data.table](https://github.com/Rdatatable/data.table/wiki/Convenience-features-of-fread#2-automatic-separator-and-skip-detection)

jimhester:
Fixed by #1172

PR #942, opened 2018-12-11
Implements a guess_delim function to guess the delimiter of a file

cimentadaj:
First rough draft of the function. Far from the actual final product and not ready to review. I will work on this slowly, porting the ideas from Sniffer as discussed in #926.


jimhester:
Thanks for filing this issue! Unfortunately, I don't think this will be merged. [vroom](https://vroom.r-lib.org/) has guessing functionality and our plan is to transition readr to use vroom as a backend in the future.

Even though we're closing this issue, we really appreciate the work, and hope you'll continue to contribute in the future :smile:

PR #949, opened 2018-12-18
Colourise types of columns when printing specs

jimhester:
I think open questions are the colors to use and if we need a readr specific way to turn this off (right now it uses `crayon::has_color()`).

I used the following to test how it looks

```r
library(dplyr)
library(nycflights13)
library(readr)
set.seed(42)
flights %>%
  transmute(
    carrier, flight, tailnum, origin, dest,
    departure = lubridate::make_datetime(year, month, day, hour, minute),
    delayed = arr_delay > 15) %>%
  sample_n(25) %>%
  write_tsv("flights.tsv")

read_tsv("flights.tsv")
```

screen shot 2018-12-18 at 2 04 37 pm

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # readr (development version)
 
+* Column specifications are now coloured when printed. This makes it easy to
+  see at a glance when a column is input as a different type then the rest.
+  Colouring can be disabled by setting `options(crayon.enabled = FALSE)`.
+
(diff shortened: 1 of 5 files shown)


jimhester:
Any thoughts on this @hadley?

hadley:
I like it; lets try it as is, and if people comment positively on it, we can eventually move it to something like `vctrs::vec_ptype_col()`

Issue #635, opened 2017-03-15
write_csv documentation examples


pgensler:
It would be really nice to have some clearer documentation with write_csv that you could do a simple example like this::

write_csv(wine, "wine.csv") where the function writes "wine.csv" to you default working directory, unless otherwise specified. Currently the documentation makes it sound like you specify only a file path, with no file name, and then no confirmation is outputted stating it was successful.

On a side note, are the examples in the documentation intended to be a MWE/ or a unit test? As I am 1000% for unit testing code, some of these examples seem more like that style, and less concise, clear and brief examples. I'd love to help with providing some, but I'm not sure if this is a roxygen thing when you make documentation for a package. Thanks for such a great package!

msalganik:
+1

jimhester:
Using tempfiles in the example avoids polluting the users' working directory when the examples are run. 
If you want to suggest more clear documentation please do! You can simply modify the comments above the functions in `R/write.R`, we can handle the roxygen side of things.

vojtechhuser:
+1 for examples. even if they polute.

pgensler:
FYI, not sure how simplistic/complex you would like the examples to be, but it might be a good idea to utilize the ```here``` package in the examples to emphasize best practices within the tidyverse?

PR #958, opened 2019-01-19
change examples to no longer use tempfiles

skaltman:
Implements changes suggested in #635 

Changes the examples so that they demonstrate write_*() behavior when you only specify a file name. Removes use of tempfile() and tempdir()


jimhester:
This is great, great job!

Issue #709, opened 2017-09-08
read_package


hadley:
Wrapper around `data()` that forces the promise. This is also a stronger parallel to loading data from a file, which is useful in intro courses.

cc @mine-cetinkaya-rundel

hadley:
Maybe package should be the second argument and it should search attached packages by default (like `data()`)

hadley:
And cc @garrettgman

jimhester:
How is this related to https://github.com/tidyverse/readr/issues/664, are they the same thing?

hadley:
Yes it's the same thing 😭

Issue #664, opened 2017-05-01
read_builtin


hadley:
Something like this:

```R
read_builtin <- function(x, package = NULL) {
  x <- data(list(x), package = package, envir = environment())
  get(x[[1]], envir = environment)
}
```

Makes `data()` consistent with the other reading functions.

mkearney:
DIBS

PR #959, opened 2019-01-19
Add read_built()

mkearney:
Creates consistent wrapper around `data()` as described conceptually in #664 and operationally #709


mkearney:
I accidentally included extraneous `write_rds()` code in the original commit (I had copied [R/rds.R](R/rds.R) to use as a reference) 😇🤷‍♂️🕺

PR #961, opened 2019-01-19
separate write_*() and format_*() examples

skaltman:
- changes format_delim() documentation so that it no longer inherits examples from write_delim()
- removes write_*() examples from format_*() docs and format_*() examples from write_*() docs


jimhester:
Thanks!

Issue #938, opened 2018-12-06
More helpful error when trying to write out data frames with list columns


rensa:
``` r
library(tidyverse)
#> Warning: package 'ggplot2' was built under R version 3.5.1
#> Warning: package 'dplyr' was built under R version 3.5.1
diamonds %>% nest(-cut) %>% write_csv('test.csv')
#> Error in stream_delim_(df, path, ...): Don't know how to handle vector of type list.
```

Created on 2018-12-06 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

It would be more useful (especially in non-interactive use) if the `write_` functions reported _which_ column it was unable to handle. My feeling is this error either happens because:

* Someone forgot to drop an intermediate list column (eg. nested source data or a list of models) before exporting;
* They accidentally used regular `map()` instead of a type-coercive variety like `map_dbl` to extract a parameter; or
* They don't understand that flat files can't store more complex list columns.

I understand that there's some discussion in #303 about potentially handling simple list columns, but given the variety of objects that end up in list columns accidentally, I imagine this problem will still exist even if that issue gets resolved. Actually implementing a PR for this might be a little out of my depth, but I'm willing to give it a go over Xmas.

PR #962, opened 2019-01-19
Error message showing column name when trying to write list column

lambdamoses:
Addresses issue #938 . Moreover, `write_delim`, and thus any function calling it, will throw an error when trying to write a list column created with base R involving calling `I()`.


jimhester:
Thanks a million!

PR #966, opened 2019-02-05
Add early check for connections with empty data

pralitp:
Add an early check if a `datasource_connection` results in an `empty_file` and just return an empty tibble if so.  Fixes tidyverse/readr#963

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # readr (development version)
 
+* `read_delimited()` function return an empty `tibble::data_frame()` rather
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks! Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

pralitp:
Done, let me know if that suffices.

jimhester:
Looks good, thanks!

PR #989, opened 2019-04-14
improve formatting at the end of the vignette

cderv:
Currently, I believe there is an issue with the formatting of the overview vignette
https://readr.tidyverse.org/articles/readr.html#available-column-specifications

This PR is just about trying to improve the formatting.


jimhester:
Thanks!

PR #1001, opened 2019-06-12
add the param version to write_rds()

shrektan:
Since serialization format 3 is the default on R 3.6.0, our users may want specify the rds serialization format they are using, due to the fact that format 3 can't be read by R version before 3.5.0.


jennybc:
For context, when faced with a related situation in `usethis::use_data()`, I chose to expose the version but default to 2. It's worth considering whether readr should do same.

https://usethis.r-lib.org/reference/use_data.html

https://github.com/r-lib/usethis/commit/12392c50d458255298c7b3e42724a7f82e74560c

jimhester:
I think that we should default to 2 as well here, as it has greater compatibility. 

Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Also could you re-run the roxygen documentation and commit the results, either `devtools::document()` or `roxygen2::roxygenize()`.

Otherwise this looks good, thanks!

shrektan:
@jimhester I've changed the default value of `version` to 2, run the docs and added an item in NEWS. Thanks. :D

shrektan:
BTW, the failed checks are caused by other reasons and not related to this PR.

image

jimhester:
A great improvement!

Thanks a million!

PR #1009, opened 2019-07-08
More helpful error when trying to write out data frames with list columns

### LINKED ISSUE #938: More helpful error when trying to write out data frames with list columns

ellessenne:
Closes #938:

``` r
library(tidyverse)
df <- diamonds %>% nest(-cut)
write_csv(df, "test.csv")
#> Error in output_column.list(x[[i]], i): Flat files can't store the list column 'data'
```

However, this approach does not work so well when multiple list columns are present:

```r
df$list <- lapply(seq(nrow(df)), rnorm, n = 3)
write_csv(df, "test.csv")
#> Error in output_column.list(x[[i]], i): Flat files can't store the list column 'data'
```

Is this behaviour acceptable or would it be better to print the names of all unsupported list columns at once?

Alessandro


R/write.R
```diff
@@ -187,26 +187,31 @@ format_tsv <- function(x, na = "NA", append = FALSE, col_names = !append, quote_
...
 
 #' @export
-output_column.POSIXt <- function(x) {
+output_column.POSIXt <- function(x, name) {
   format(x, "%Y-%m-%dT%H:%M:%OSZ", tz = "UTC", justify = "none")
 }
 
+#' @export
+output_column.list <- function(x, name) {
+  stop(paste0("Flat files can't store the list column '", name, "'"))
```
hadley: I think this should use `` ` `` rather than `'`, and you need a `call. = FALSE` on the end. Also `stop()` automatically pastes together it's inputs, you can remove the `paste()` if you want.

tests/testthat/test-write.R
```diff
@@ -163,3 +163,11 @@ test_that("hms NAs are written without padding (#930)", {
   df <- data.frame(x = hms::as.hms(c(NA, 34.234)))
   expect_equal(format_tsv(df), "x\nNA\n00:00:34.234\n")
 })
+
+test_that("More helpful error when trying to write out data frames with list columns (#938)", {
+  df <- tibble::tibble(id = seq(11))
```
hadley: I'd make this a little simpler: `df <- tibble(x = 1, y = list(1))`

tests/testthat/test-write.R
```diff
@@ -163,3 +163,11 @@ test_that("hms NAs are written without padding (#930)", {
...
+  df$list <- lapply(X = seq_along(df$id), FUN = rnorm, n = 3)
+  filename <- file.path(tempdir(), "test_list.csv")
+  on.exit(unlink(filename))
+  expect_error(write_csv(x = df, path = filename), "Flat files can't store the list column '")
```
hadley: I'd just do `write_csv(df, tempfile())` — you don't need to clean up since this won't read a file.

ellessenne:
Thanks @hadley, I push some new changes following your suggestions!

hadley:
Thanks @ellessenne

PR #1024, opened 2019-08-26
Convert GitHub URLs to raw links

### LINKED ISSUE #1023: Automatically convert github links to follow the raw link???
Maybe readr should detect GitHub links and automatically convert them to the raw links, so you don't have to copy paste the raw link, e.g. 
```r
# This
readr::read_csv("https://github.com/zonination/pokemon-chart/blob/master/chart.csv")

# rather than this
readr::read_csv("https://raw.githubusercontent.com/zonination/pokemon-chart/master/chart.csv")
```

Many people may not be aware of the distinction, and not understand why readr is unable to parse the former, as the URL makes it look like CSV even though it is actually HTML.

jimhester:
This is not done, would like to add some tests, but here is a simple implementation.

Fixes #1023

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -126,6 +126,10 @@ standardise_path <- function(path, input = TRUE) {
     return(path)
 
   if (is_url(path)) {
+    if (is_github_url(path)) {
+      path <- convert_github_url(path)
+    }
+
     if (requireNamespace("curl", quietly = TRUE)) {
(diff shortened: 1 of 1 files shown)


jimhester:
Hi past self, this is a good idea and would be useful in some cases, but for now I think it best if we close it.

PR #1034, opened 2019-10-09
spelling

wibeasley:
Nothing major in this PR.  It just reduces the number of "Potential spelling errors" listed when checking, so hopefully new mistakes are easier to catch.

PR #1039, opened 2019-10-17
update wordlist for spellcheck

wibeasley:
follows @jeroen's & @jimhester's advice in #1034 and runs `spelling::update_wordlist()`.

It makes me a little cautious to remove words like 'RStudio', but I guess they're built into the base dictionary now.

That's a helpful package that I wasn't aware of.  I should do some research and find out who the two developers are.


jimhester:
I think spelling more likely just no longer tries to spell check capitalized words, assuming they are proper nouns.

Thanks for the PRs!

PR #1045, opened 2019-11-07
Add `use_tidy_github()` docs and LICENSE.md

batpigandme:
cc @hadley 
Like ggplot2, this one was GPL2, so I just added the `LICENSE.md`, and added that to the `.Rbuildignore`.


jimhester:
Thanks!

PR #1048, opened 2019-11-18
`write_csv2` and `format_csv2` no longer pad number columns with whitespace

keesdeschepper:
`write_csv2` and `format_csv2` no longer add leading whitespace to number columns with numbers of different character lengths. Fixes tidyverse#1046


jimhester:
Great work!

Thanks!! You are da bomb!

PR #1070, opened 2020-01-31
Added new examples to parse_number

### LINKED ISSUE #1057: Better examples for parse_number()
The second example is confusing: https://github.com/tidyverse/readr/issues/1007

And it for such an useful function, it would be nice to have a few more examples of realistic usage.

cdhowe:
Fixes #1057

### DIFF
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -138,8 +138,23 @@ col_skip <- function() {
 #' @family parsers
 #' @export
 #' @examples
-#' parse_number("$1000")
+#' ## These all return 1000
+#' parse_number("$1,000") ## leading $ and grouping character , ignored
+#' parse_number("£1000")  ## leading non-numeric £ ignored (locale doesn't matter)
+#' parse_number("AA1000") ## all leading non-numeric characters ignored
+#'
+#'
+#' ## These all return 1234567.78
+#' ## If you don't see the decimal digits, try options(digits=10)
 #' parse_number("1,234,567.78")
+#' ## explicit locale specifying European grouping and decimal marks
+#' parse_number("1.234.567,78", locale = locale(decimal_mark = ",", grouping_mark = "."))
+#' ## SI/ISO 31-0 standard spaces for number grouping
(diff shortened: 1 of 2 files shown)


R/collectors.R
```diff
@@ -138,8 +138,23 @@ col_skip <- function() {
 #' @family parsers
 #' @export
 #' @examples
-#' parse_number("$1000")
+#' ## These all return 1000
+#' parse_number("$1,000") ## leading $ and grouping character , ignored
+#' parse_number("£1000")  ## leading non-numeric £ ignored (locale doesn't matter)
```
jennybc: I would avoid adding a challenging character like this.

@jimhester what do you think?

R/collectors.R
```diff
@@ -138,8 +138,23 @@ col_skip <- function() {
...
+#' parse_number("AA1000") ## all leading non-numeric characters ignored
+#'
+#'
+#' ## These all return 1234567.78
+#' ## If you don't see the decimal digits, try options(digits=10)
```
jennybc: I did have to set options to see this the promised result.

So, if we invite people to change their options ... we need to show them `op <- options(digits = 10)` followed by `options(op)` (restoring the original settings). So should we just do that?

Again this is @jimhester's call.

R/collectors.R
jimhester: Yeah I think that is right, it actually is a great example, but unfortunately non-ASCII characters often cause us problems in examples on CRAN

jimhester: @cdhowe I think we have to remove this example unfortunately

R/collectors.R
jimhester: I am somewhat wary of setting options in examples in case the examples fail. I think just documenting this is OK in this case.

cdhowe:
Changed examples to address comments -- specifically dropping the example with the pound sign and shortening numbers to we don't need to adjust digits.

jimhester:
Great, thanks @cdhowe!

PR #1089, opened 2020-04-12
Minor errors in guess_parser docs.

dmurdoch:
The previous text says "we don't guess that currencies are numbers, even though we can parse them", and then the example didn't try to parse a currency value.  The next text has "The" when it should have "There".

### DIFF
--- a/vignettes/readr.Rmd
+++ b/vignettes/readr.Rmd
@@ -122,10 +122,10 @@ The guessing policies are described in the documentation for the individual func
 
 ```{r}
 guess_parser("$1,234")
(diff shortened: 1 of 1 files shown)


dmurdoch:
Checks failed, because of problems in `write.R`:  the `eol` description forgot to put backticks around the newline strings, so the "\n" and "\r\n" make it into the Rd without being properly wrapped in \code{}.  I'd submit a PR to fix this too, but I can't even run roxygenize right now, because of some compile issue that's probably an out of date header on my system.

jimhester:
Thanks!

PR #1091, opened 2020-04-13
Typo:  unique names have a suffix, not a prefix

dmurdoch:


jimhester:
Yep, clearly wrong, thanks Duncan!

PR #1110, opened 2020-06-13
Change path argument to file in write_csv

### LINKED ISSUE #964: file vs path argument to read_* and write_*
I can not remember if `read_*`  or `write_*` requires the path argument, and which of those two needs file.
How about having the same argument name, either file or path, for read_ and write_?

brianrice2:
The `read_*` and `write_*` functions use different parameters of
file vs path, which can be confusing. Replaced path argument with file
in `write.R`, similar to existing structure in `read_delim.R`.

Also updated accompanying unit test in `test-write.R`. Results
remain 759 OK / 7 Fail / 1 Warning.

Fixes #964.


jimhester:
Thank you!

PR #1112, opened 2020-06-17
Add `cols_regex()` (preliminary proposal)

salim-b:
This PR adds the function **`cols_regex()`** to create a columns specification using regular expression matching.

_This is only a **preliminary proposal** since I couldn't figure out a way to automatically determine the column names of the data needed to match against. Thus one needs to manually provide those in `cols_regex(.column_names = ...)`, which is obviously not very convenient._

The basic idea behind `cols_regex()` is that you can provide a single regular expression per desired "column type" (more precisely per column specification object, e.g. `col_integer()` or its string abbreviation, e.g. `"i"`) instead of having to write out all column names (which you simply might not know beforehand when parsing results from certain web services for example).

I've included a real-life example in `inst/extdata/ch_communes_snapshot.csv`. To parse its 34 columns each with the proper type:

```r
raw_data <- datasource(file = system.file("extdata/ch_communes_snapshot.csv",
                                          package = "readr"))

col_names <- readr:::guess_header(datasource = raw_data,
                                  tokenizer = tokenizer_csv())$header

read_csv(
  file = raw_data,
  col_types = cols_regex("(Name|_Title|_Text|^ABBREV)" = "c",
                         "^(MutationDate|ValidFrom|ValidTo)$" = col_date(format = "%d.%m.%Y"),
                         .default = "i",
                         .col_names = col_names)
)
```

Now as you can see, I have to manually extract the column names from the CSV file in order to feed it to `cols_regex()`. Ideally, this wouldn't be necessary since `read_csv()` does the same header parsing eventually. But as far as I understand R internals, it's not possible to just carry over `read_csv`'s arguments (`file`, `locale`, `skip` and `skip_empty_rows`) to `cols_regex()`, right?

And even if so, two downsides would remain:
1. We would still need to provide the proper tokenizer to `guess_header()`.
2. It would still result in parsing the CSV header _twice_ which doesn't seem right.

So I suppose a more user-friendly API would require some internal architectural changes...

Nevertheless, **I'd be very happy about feedback from people more familiar with readr's internals**! Maybe there's even a more elegant way to accomplish what I want (specify column types by regular expressions matching the column names)?

<sub>Of course I'm aware that I could simply parse the whole CSV as character only and then change types afterwards using dplyr or whatever, but that's really not the point. 😄</sub>


jimhester:
> Of course I'm aware that I could simply parse the whole CSV as character only and then change types afterwards using dplyr or whatever, but that's really not the point. 😄

FWIW no need for dplyr to do that particular thing, the idiom is as follows.

```r
chr_data <- readr::read_csv(col_types = list(.default = "c"))
data <- readr::type_convert(chr_data)
```

I haven't looked at the rest of the code, will do so tomorrow, but in general my feeling is this is not an avenue we want to go down.

salim-b:
> FWIW no need for dplyr to do that particular thing, the idiom is as follows.
> 
> ```r
> chr_data <- readr::read_csv(col_types = list(.default = "c"))
> data <- readr::type_convert(chr_data)
> ```

Thanks for the hint to `readr::type_convert()`. But it doesn't allow what I want to achieve, i.e. it requires the same column spec object as `read_csv()` (provided as arg `col_types`) _if_ the column types should not be guessed (guessing is error-prone, even if it's based on an imputation using all rows as `type_convert()` does, because _all_ rows is a pretty relative term in case of results from web services).

Therefore I think having a native `cols_regex()` function would still make sense :)

salim-b:
@jimhester , have you had a chance to take a closer look at my PR in the meantime?

IMHO it would be awesome if an improved version of my `cols_regex()` function could be added to readr. Improved in the sense that the header parsing would be done only once (internally) and the user wouldn't need to supply the `.col_names` argument (as described in the PR message).

I'd be happy to document possible usage examples in a short vignette if desired, just let me know.

jimhester:
Thanks for working on this pull request! Unfortunately, while I do see the power of a function like this I do not think it is something we will be adding to readr.

Even though we're closing this issue, we really appreciate the feedback, and hope you'll continue to contribute in the future :smile:

salim-b:
Bummer to hear that, but thanks for the kind words! :)

Should anyone stumble upon this issue and would like to use `cols_regex()`, it's part of my [pal](https://gitlab.com/salim_b/r/pkgs/pal) package.

PR #1142, opened 2020-11-01
Fix escaping of % on col_date()/col_datetime()'s doc

yutannihilation:
It seems this part of the doc is now broken. Maybe this is forgotten to be unescaped when migrated to markdown format?

https://readr.tidyverse.org/reference/parse_datetime.html#format-specification:
image


jimhester:
Thanks! You are da bomb!

PR #1150, opened 2020-11-17
Add minimal lifecycle version

maelle:
since readr uses `is_present()` https://github.com/r-lib/lifecycle/blob/master/NEWS.md#lifecycle-020


jimhester:
Thanks!

PR #1152, opened 2020-11-18
fix read_delim() fails when sample of parsing problems contains non-ASCII characters

### LINKED ISSUE #1136: read_delim() fails when sample of parsing problems contains non-ASCII characters
When a non-ASCII file is read using `read_delim()` or its friends *and* there are parsing failures *and* the text used by `warn_problems()` contains non-ASCII characters, the following error is returned:

```
## Error in nchar(x) : invalid multibyte string, element 3
```

This is quite hard to debug because the error makes it seem like the encoding was set incorrectly or the file has problems. (In my case it only occurred with 2 out of about 6000 CSV files read using `map_dfr()` even though all contained non-ASCII text.)

In fact the error occurs in `warn_problems()`: 

https://github.com/tidyverse/readr/blob/05890c3b493458ad4699654ceed58d2515939e94/R/problems.R#L83

where `nchar()` throws an error if one of the problems "sampled" for display contains a non-UTF-8 string. The data is in fact read correctly inside `read_delimited()`; the call only fails once `warn_problems()` is called.

Once I supplied the column specification and the non-ASCII text no longer appeared in the list of reported problems, the error no longer occurred (see reprex).

Possibly related to #1111 where the same error is thrown by `parse_number()`; this may have a similar root cause though I haven't investigated specifically.

Reprex:

```r
tf <- tempfile()
download.file("https://vdp.cuzk.cz/vymenny_format/csv/20200930_OB_547999_ADR.csv.zip", destfile = tf)
read_csv2(tf, locale = locale(encoding = "Windows-1250"))

## Error in nchar(x) : invalid multibyte string, element 3
```

Inspecting this in the debugger, the `probs_f` variable used inside `warn_problems()` does indeed contain non-ASCII text.

This works as expected:

```
read_csv2(tf, locale = locale(encoding = "Windows-1250"),
          col_types = cols(.default = "c"))

```

The file referenced in the reprex above was correctly read using `read.csv2` also using the "WINDOWS-1250" encoding. Likewise it can be read in python using the pandas CSV reader with the same encoding.

System:

Mac OS
R 4.0.2
readr 1.4.0 from CRAN

hidekoji:
fix #1136 

### Root Cause

In `warn_problems`, there was a `nchar()` function call but this does not work when the value contains non-ascii characters.

```R
  width <- vapply(probs_f, function(x) max(nchar(x)), integer(1))
```

To make it work with non-ascii characters, it needs to call `encodeString` before passing the value to `nchar`.

### Fix

Apply `encodeString` so that it won't fail with Invalid multibyte string, element 3

### How to confirm the fix

You can confirm the fix by  below examples:

Case 1. (Czech)

```r
tf <- tempfile()
download.file("https://vdp.cuzk.cz/vymenny_format/csv/20200930_OB_547999_ADR.csv.zip", destfile = tf)
read_csv2(tf, locale = locale(encoding = "Windows-1250"))
 ```

Case 2. (Japanese)

```r
readr::read_delim("https://www.dropbox.com/s/aji9zb5qqu76hn8/13_2018.csv?dl=1" , ",", 
                  quote = "\"", skip = 0 , col_names = TRUE , na = c('','NA') , 
                  locale=readr::locale(encoding = "SHIFT_JIS", 
                                       decimal_mark = ".", tz = "America/Los_Angeles", grouping_mark = "," ), trim_ws = TRUE , progress = TRUE)
```


hidekoji:
Any update on this PR?

jimhester:
It looks fine thanks! Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jimhester:
And could you add a test for this case, so we avoid future regressions?

hidekoji:
hmm, R-CMD-check failed on ubuntu-16.0.4 (3.3) and it complains about the package size and not sure why the size is different with ubuntu-16.0.4 (3.3)...

image

Could you please help me to resolve the check?

jimhester:
The error seems unrelated to this PR, and was actually a segfault when running the vignettes, not due to the package size. In any case thanks for working on this!

Thanks!

PR #1153, opened 2020-11-18
Fixes collectors documentation (e.g. parse_date).

adamroyjones:
When using explicit LaTeX-like "itemize" and "enumerate" environments, special characters require escaping. Currently, the documentation for collectors does *not* escape its special characters. This leads to issues such as the following:

maim-2020-11-18T20-09-19

This commit addresses this issue by instead using Markdown within documentation blocks. This has been tested locally by building the documentation with `pkgdown::build_site()` and verifying that "parse_datetime.html" was rendered in the expected way. The screenshot below illustrates the output.

maim-2020-11-18T20-10-05


adamroyjones:
(I'll reopen it shortly; there was a mistake in my commit message.)

PR #1154, opened 2020-11-18
Fixes collectors documentation (e.g. parse_date).

adamroyjones:
When using explicit LaTeX-like "itemize" and "enumerate" environments, special characters require escaping. Currently, the documentation for collectors does *not* escape its special characters. This leads to issues such as the following:

maim-2020-11-18T20-09-19

This commit addresses this issue by instead using Markdown within documentation blocks. This has been tested locally by building the documentation with `pkgdown::build_site()` and verifying that "parse_datetime.html" was rendered in the expected way. The screenshot below illustrates the output.

maim-2020-11-18T20-10-05

### DIFF
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -260,55 +260,47 @@ col_factor <- function(levels = NULL, ordered = FALSE, include_na = FALSE) {
 #' `readr` uses a format specification similar to [strptime()].
 #' There are three types of element:
 #'
-#' \enumerate{
-#'   \item Date components are specified with "%" followed by a letter.
-#'     For example "%Y" matches a 4 digit year, "%m", matches a 2 digit
-#'     month and "%d" matches a 2 digit day. Month and day default to `1`,
-#'     (i.e. Jan 1st) if not present, for example if only a year is given.
-#'   \item Whitespace is any sequence of zero or more whitespace characters.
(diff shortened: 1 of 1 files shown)


adamroyjones:
(I'm sorry for the duplicated pull request. I don't use GitHub all that often and thought that I could delete the other pull request, but I can't, which is just super.)

jimhester:
Thanks, however this is already fixed in https://github.com/tidyverse/readr/pull/1142 as you can see in the dev version of the documentation. https://readr.tidyverse.org/dev/reference/parse_datetime.html

adamroyjones:
> Thanks, however this is already fixed in #1142 as you can see in the dev version of the documentation. https://readr.tidyverse.org/dev/reference/parse_datetime.html

I didn't see that (and didn't know that there was a dev version of the documentation)—thanks!

The changes aren't _entirely_ vacuous, however, though they're reduced to a triviality: some of the documentation (e.g. around `col_types`) is formatted with Markdown, and some is not (see screenshot below).

Screenshot 2020-11-19 at 14 50 47

If this is an unwanted change, feel free to close, delete, and purge it into oblivious; if you want the change but with reworded commits and branches, I'm happy to update or reraise the request.

jimhester:
Thanks!

PR #1172, opened 2021-01-13
Converting readr to use vroom as the default backend.

jimhester:


jimhester:
Differences from current readr behavior.

- comments within data are not supported. This is not really possible to support without changing what the vroom index stores.
- skipped lines within data are not supported. This is not really possible to support without changing what the vroom index stores.
  - This implies the `skip_empty_rows()` argument is deprecated.
- treating length > 1 character vectors as literal data is not supported. vroom treats these as multiple files to read.
- unnamed `col_types` must equal the number of columns in the file, in readr this is a warning. (https://github.com/r-lib/vroom/issues/296)
- readr names the columns in `problems()`, vroom uses the column index.
- row number in `problems()`, should it be based on the _file_ line number or the _data_ row. https://github.com/r-lib/vroom/issues/298

BEFH:
> unnamed col_types must equal the number of columns in the file, in readr this is a warning. (r-lib/vroom#296)

This is a huge change in behavior and extremely problematic for my existing code base. Will there at least be an error rather than a warning so that my scripts will not silently fail?

Also, is there no way to internally change the input to a `cols_only()` specification for `col_types()`?

PR #1194, opened 2021-04-13
Use the clock package to convert date times

jimhester:
A port of https://github.com/r-lib/vroom/pull/322 to readr


jimhester:
The code here is mostly the same as in vroom, but it would still be useful if you did a quick look over @DavisVaughan.

DavisVaughan (approved):
LGTM!

PR #1232, opened 2021-07-20
include `*_chunked()` functions in pkgdown TOC

wibeasley:
When I built it locally on a Windows machine, it had two errors (unrelated to pkgdown) that looked like a Windows/Linux line ending difference

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -55,6 +55,7 @@ reference:
     line is divided up into fields.
   contents:
   - read_delim
+  - read_delim_chunked
   - read_fwf
   - read_log
   - read_table
@@ -119,6 +120,7 @@ reference:
     the file, and how each line is divided up into fields.
   contents:
   - melt_delim
+  - melt_delim_chunked
   - melt_fwf
   - melt_table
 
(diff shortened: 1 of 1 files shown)


jimhester:
I actually deliberately removed these functions from the pkgdown documentation in 696ddc2665df98.

These functions have always been mostly experimental and also rely on the 1st edition parser, they may go away some time in the future.

wibeasley:
Gotcha. I'll close the PR.  FWIW, that chunked function was really nice for an isolated case I had a few months ago when someone dumped a single 3GB file that I needed to tweak before uploading into real database.  It was nice how the chunking function bit off a little at a time and delegated it to the callback function.  R never needed to keep the whole thing in ram.

But at least in my experience, that scenario comes up rarely …certainly not enough to justify continual support in a popular package.  I guess if I had to implement it without the native chunking function, I would initially get a row count (maybe with the new readr/vroom option of a single column).  And then loop through using `skip` and `n_max` of [`readr::read_csv()`](https://readr.tidyverse.org/reference/read_delim.html).

------

And congrats for getting vroom mature enough to pull into readr.  I've been using it for 2 years for big files and like it a lot.

PR #1239, opened 2021-07-21
update documentation to reflect change to `read_delim` API

### LINKED ISSUE #1238: `read_delim` documentation for `file` arg is inaccurate
From the docs for the `file` argument: "Literal data is most useful for examples and tests. It must contain at least one new line to be recognised as data (instead of a path) or be a vector of greater than length 1."

It fails to note that in the new edition, a multi-element vector will now be interpreted as a list of paths unless you wrap it with `I()`.

tarheel:
fixes #1238

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -9,9 +9,9 @@
 #'    downloaded. Remote gz files can also be automatically downloaded and
 #'    decompressed.
 #'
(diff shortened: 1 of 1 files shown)


jimhester:
Thanks a bunch!

PR #1332, opened 2021-11-22
Add support for show_col_types for edition 1 parser

### LINKED ISSUE #1331: options(readr.show_col_types=FALSE) is not work on read_table()
options(readr.show_col_types=FALSE) is not work on read_table() and read_table is not allow set the parameter in the function for example:
read_table(show_col_types=FALSE)

jimhester:
`read_table()` is one of the functions that currently doesn't have a
vroom equivalent, so is still using the first edition parser. When we
added support for `show_col_types` we didn't port that back to the first
edition parser, so it was missing in `read_table()` and the
`read_delim_chunked()` functions.

Fixes #1331

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
 * Jenny Bryan is now the maintainer.
 
 * `read_table()` and edition 1 parsers gain support for `show_col_types()` (#1331)
+* `read_table()`, `read_log()`, and `read_delim_chunked()` (and friends) gain the `show_col_types` argument found elsewhere. All `read_*()` functions now respect the `show_col_types` argument or option even when using the first edition parsing engine (#1331).
 * Fix buffer overflow when trying to parse an integer from a field that is over 64 characters long (#1326)
 
 # readr 2.1.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 
 * Jenny Bryan is now the maintainer.
 
+* `read_table()` and edition 1 parsers gain support for `show_col_types()` (#1331)
 * Fix buffer overflow when trying to parse an integer from a field that is over 64 characters long (#1326)
 
 # readr 2.1.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,11 +1,11 @@
 # readr (development version)
 
+* `read_table()`, `read_log()`, and `read_delim_chunked()` (and friends) gain the `show_col_types` argument found elsewhere. All `read_*()` functions now respect the `show_col_types` argument or option even when using the first edition parsing engine (#1331).
+
 # readr 2.1.1
 
 * Jenny Bryan is now the maintainer.
 
-* `read_table()` and edition 1 parsers gain support for `show_col_types()` (#1331)
-* `read_table()`, `read_log()`, and `read_delim_chunked()` (and friends) gain the `show_col_types` argument found elsewhere. All `read_*()` functions now respect the `show_col_types` argument or option even when using the first edition parsing engine (#1331).
 * Fix buffer overflow when trying to parse an integer from a field that is over 64 characters long (#1326)
 
 # readr 2.1.0
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -390,13 +390,6 @@ read_tokens <- function(data, tokenizer, col_specs, col_names, locale_, n_max, p
   read_tokens_(data, tokenizer, col_specs, col_names, locale_, n_max, progress)
 }
 
-should_show_col_types <- function(has_col_types, show_col_types) {
-  if (is.null(show_col_types)) {
-    return(isTRUE(!has_col_types))
-  }
-  isTRUE(show_col_types)
-}
-
 read_delimited <- function(file, tokenizer, col_names = TRUE, col_types = NULL,
                            locale = default_locale(), skip = 0, skip_empty_rows = TRUE, skip_quote = TRUE,
                            comment = "", n_max = Inf, guess_max = min(1000, n_max), progress = show_progress(),
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -427,7 +427,10 @@ read_delimited <- function(file, tokenizer, col_names = TRUE, col_types = NULL,
 
   has_col_types <- !is.null(col_types)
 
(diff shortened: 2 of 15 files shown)


jimhester:
I didn't add a test for this explicitly, but I can if you think it would be useful.

R/read_delim.R
```diff
@@ -420,7 +432,9 @@ read_delimited <- function(file, tokenizer, col_names = TRUE, col_types = NULL,
 
   ds <- datasource(data, skip = spec$skip, skip_empty_rows = skip_empty_rows, comment = comment, skip_quote = skip_quote)
 
-  if (is.null(col_types) && !inherits(ds, "source_string") && !is_testing()) {
+  has_col_types <- !is.null(col_types)
+
+  if (((is.null(show_col_types) && !has_col_types) || isTRUE(show_col_types)) && !inherits(ds, "source_string") && !is_testing()) {
```
jimhester, line 437: This logic is a bit confusing, basically if `show_col_types` is `NULL`, which is the default then we only want to show the column types if they are not explicitly set, or if we are running the tests or getting data from a string.

Suggestions welcome on ways to make it more understandable.

src/cpp11.cpp
```diff
@@ -197,34 +198,6 @@ extern "C" SEXP _readr_write_file_raw_(SEXP x, SEXP connection) {
 }
 
 extern "C" {
-/* .Call calls */
```
jimhester, line 200: These changes in cpp11.cpp are just due to using a newer version of cpp11 since the last time this was run, they can be safely ignored.

R/read_delim.R
```diff
@@ -388,7 +392,8 @@ read_tokens <- function(data, tokenizer, col_specs, col_names, locale_, n_max, p
 
 read_delimited <- function(file, tokenizer, col_names = TRUE, col_types = NULL,
                            locale = default_locale(), skip = 0, skip_empty_rows = TRUE, skip_quote = TRUE,
-                           comment = "", n_max = Inf, guess_max = min(1000, n_max), progress = show_progress()) {
+                           comment = "", n_max = Inf, guess_max = min(1000, n_max), progress = show_progress(),
+                           show_col_types = should_show_col_types()) {
```
jennybc, line 396:
```suggestion
                           show_col_types = should_show_types()) {
```
```suggestion
                           show_col_types = should_show_col_types()) {
```

jennybc (commented):
LGTM except my 1 suggestion re: a function-o.

I think it would be good to add some tests. Snapshot tests? I see readr is using testthat 3e, in terms of what DESCRIPTION says, but I haven't looked around at the actual usage.

Let me know if you want me to extend this PR with some tests.

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # readr (development version)
 
+* `read_table()` and edition 1 parsers gain support for `show_col_types()` (#1331)
```
jennybc, line 3:
```suggestion
* `read_table()`, `read_log()`, and `read_delim_chunked()` gain the `show_col_types` argument found elsewhere. All `read_*()` functions now respect the `show_col_types` argument or option even when using the first edition parsing engine (#1331).
```

R/read_delim.R
jennybc: It seems like the presence of `is_testing()` will be an obstacle to snapshot testing, which would be nice for the col spec stuff, so we might want to refactor a bit. If I end up adding tests, I could work on that.

jimhester:
Thanks for the fixes, and sure we can add some snapshot tests, you are welcome to refactor is_testing as needed to so you can do it.

R/utils.R
```diff
@@ -27,12 +27,22 @@ show_progress <- function() {
...
-#' - They are disabled by setting `options(readr.show_col_types = FALSE)`
-#' - The column types are supplied with the `col_types` argument.
+#' Wrapper around `getOption("readr.show_col_types")` that implements some fall
+#' back logic if the option is unset. This returns:
+#' * `TRUE` if the option is set to `TRUE`
+#' * `FALSE` if the option is set to `FALSE`
+#' * `FALSE` if the option is unset and we appear to be running tests
+#' * `NULL` otherwise, in which case the caller determines whether to show
+#'   column types based on context, e.g. whether `show_col_types` or actual
+#'   `col_types` were explicitly specified
```
jennybc, line 37: I reworded this to focus more closely on what this helper does. I still include some words about downstream use, in the `NULL` case.

R/utils.R
```diff
@@ -27,12 +27,22 @@ show_progress <- function() {
...
 #' @export
 should_show_types <- function() {
-  if (identical(getOption("readr.show_col_types", TRUE), FALSE)) {
+  opt <- getOption("readr.show_col_types", NA)
+  if (isTRUE(opt)) {
+    TRUE
```
jennybc, line 42: This is new. But this just feels more natural and expected to me that if the option is `TRUE` or `FALSE`, that value is returned.

R/utils.R
```diff
@@ -27,12 +27,22 @@ show_progress <- function() {
...
+  } else if (identical(opt, FALSE)) {
+    FALSE
+  } else if (is.na(opt) && is_testing()) {
```
jennybc, line 45: I moved the `is_testing()` matter here.

tests/testthat/setup.R
```diff
@@ -1,4 +1,3 @@
 pre_test_options <- options(
-  readr.show_col_types = FALSE,
```
jennybc, line 2: I prefer to leave this unset and let `should_show_types()` fill in `FALSE` in the testing context.

R/read_delim.R
```diff
@@ -420,7 +425,12 @@ read_delimited <- function(file, tokenizer, col_names = TRUE, col_types = NULL,
 
   ds <- datasource(data, skip = spec$skip, skip_empty_rows = skip_empty_rows, comment = comment, skip_quote = skip_quote)
 
-  if (is.null(col_types) && !inherits(ds, "source_string") && !is_testing()) {
+  has_col_types <- !is.null(col_types)
+
+  if (
+    ((is.null(show_col_types) && !has_col_types) || isTRUE(show_col_types)) &&
+    !inherits(ds, "source_string")
+  ) {
```
jennybc, line 433: I moved the `is_testing()` matter into `should_show_types()`.

PR #1347, opened 2022-01-08
Test failure if file doesn't create

MichaelChirico:
Ran into a case where "foo.csv" wasn't created, then the test failed a bit confusingly as "foo.csv not found"

Adding the test here would make it a bit clearer how to debug.

### DIFF
--- a/tests/testthat/test-read-csv.R
+++ b/tests/testthat/test-read-csv.R
@@ -211,9 +211,10 @@ test_that("empty file with col_names and col_types creates correct columns", {
 })
 
 test_that("empty file returns an empty tibble", {
-  file.create("foo.csv")
+  expect_true(file.create("foo.csv"))
+  on.exit(file.remove("foo.csv"))
+
   expect_equal(read_csv("foo.csv")[], tibble::tibble())
-  file.remove("foo.csv")
 })
 
 


jennybc:
Just curious: why didn't the file get created?

I'm asking because, in full generality, I guess this sort of thing could affect many tests.

jennybc:
Sidebar (which I can deal with): but this file should really be created below the temp directory, now that I look at this.

jennybc:
I decided to just make this test work more like others (and like it should), i.e. write below temp dir and clean up after itself.

I don't want add an expectation for file creation here specifically. Why here and not for the other ~30 files that are created during tests? It seems likely that writing into the package directory might have been the actual cause for the failure you saw anyway.

MichaelChirico:
> I decided to just make this test work more like others (and like it should), i.e. write below temp dir and clean up after itself.
> 
> I don't want add an expectation for file creation here specifically. Why here and not for the other ~30 files that are created during tests? It seems likely that writing into the package directory might have been the actual cause for the failure you saw anyway.

yep that's right. originally due to a google-specific issue where tests are not always run in a writeable dir (which differs from CRAN), but my thinking was that such write failure _could_ happen on any system -- maybe the disk is full or the OS just has a hiccup.

anyway thanks for the fix!

PR #1348, opened 2022-01-08
Add a version dependency on testthat

MichaelChirico:
`variant=` argument of `expect_snapshot` was added then:

https://github.com/tidyverse/readr/blob/5a098243c026edb46d8c1cbb2f9f3a69297aac10/tests/testthat/test-read-csv.R#L233

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -60,7 +60,7 @@ Suggests:
     rmarkdown,
     spelling,
     stringi,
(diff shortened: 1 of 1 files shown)


jennybc:
Minimum versions in Suggests aren't really supported/enforced as with Imports, but yes it can be good to specify if someone has actually bumped their shins on it.

PR #1353, opened 2022-01-19
add vignette for column type guessing

### LINKED ISSUE #1352: Initiate a new vignette on column type guessing and specification
I think column type guessing and specification generates enough user interest and questions to get a dedicated vignette.

The first topic to document here is "how do I tell readr to use all of the data to guess column types"?

* Document differences between 1st and 2nd edition parsing engines
* Encourage users to abandon type guessing and, instead, provide column types, as work matures past the "exploration & discovery" phase

sbearrows:
Closes #1352

Just a starting point

I didn't create a table because I wasn't sure it would work here


vignettes/col_types.Rmd
```diff
@@ -0,0 +1,51 @@
...
+title: "Column type guessing"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Column type guessing}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+library(readr)
```
jennybc, line 11: I have a personal rule that all code critical to making things work should be visible. Let's pull the `library(readr)` call out into a user-visible chunk.

`usethis::use_vignette()` will initiate a new vignette with the recommended two "starter" chunks. You could call it once to create a dummy vignette, see what it does, emulate that here (and delete the dummy vignette).

jennybc (commented):
This definitely has the right content.

Before we meet today, can you respond to the one inline comment and implement the mechanical changes described below? Then we can do more review / word-smithing "live".

* Use one sentence per line. This has huge benefits for git diffs and makes it much easier for someone to make small suggestions to a PR.
* Hoist the second edition stuff to the top. Most people are oblivious to the edition thing, so we should address the default case first.

vignettes/col_types.Rmd
sbearrows: I did `usethis::use_vignette()` but then I saw the other vignettes had `library(readr)` in the `include = FALSE` code chunk at the top so I moved it up there. I'll move it down so readers can see it.

jennybc:
As discussed, I worked a lot on the prose and structure.

But I didn't finish 😬

I've sketched what I think is a good example dataset that stress-tests the column type guessing.

Will you pick it up from here and create some chunks in the last section / bullet list showing how the various methods play out?

jennybc:
Writing TODOs as a checklist

- [ ] Look at usage of the words "lines", "rows", "observations", etc. and see if we can standardize the vocab more.
- [ ] The first (or some early) mention of vroom should probably be `vroom::vroom()` in order to get the automatic hyperlink.
- [ ] Take the example code at the end and divide it up over a few chunks (set up + the various demos) inserted into the logical place in the prose.

sbearrows:
- [ ] Look at usage of the words "lines", "rows", "observations", etc. and see if we can standardize the vocab more.
 - [x] The first (or some early) mention of vroom should probably be vroom::vroom() in order to get the automatic hyperlink.
- [x]  Take the example code at the end and divide it up over a few chunks (set up + the various demos) inserted into the logical place in the prose.
- [x]  Link to this vignette from other obvious places, like the help for the guess_max parameter. In a roxygen comment, refer to the vignette via `vignette("column-types")` to get an automatic hyperlink in some contexts.
 - [x] In an html comment (there are some existing examples), record any/some of the "classic" readr issues where this matter has come up before. Especially for the type_convert() approach.

I think the last two points might need more links + comments but other than that I just need to do the first one.

PR #1359, opened 2022-01-25
Update homepage

sbearrows:
I've changed the tone for column guessing and automation. I still need to add in information about `readr` defaulting to `vroom::vroom()` for writing and parsing as of v 2.0.0 but there were a few different places it could go. I was thinking it could go under Alternatives. I've also added in the link to the new vignette for column type guessing.


jennybc:
I will come back to this tomorrow to make some suggestions (maybe literally, i.e. the mini PRs to your PR). But overall this is on the right track and it's great that you went ahead and converted to the one-sentence-per-line format.

jennybc:
So you should push assertively in the morning before I really come online.

You can lift some sentence of inspiration re: the vroom situation from the recently added vignette.

In code, we like to DRY things out. But with docs, it's often a good idea to repeat things. In this case, I think it's OK to repeat some content about the deal with 1st edition vs 2nd.

jennybc:
Also: in the same way that you have to run `devtools::document()` to make sure that changes to roxygen comments below `R/` are reflected in re-built files below `man/`, we have to regularly run `devtools::build_readme()` to remake `README.md` from `README.Rmd`.

If git is not fighting you re: committing `README.Rmd` without also committing `README.md`, that suggests the repo was not setup with `create_from_github()`, which sets up a Git hook around this. This is a minor matter, but one we should sort out.

README.Rmd
```diff
@@ -129,6 +129,20 @@ Compared to the corresponding base functions, readr functions:
   automatically guessing the delimiter, whether or not the file has a
   header, and how many lines to skip.
 
+
+## Versions
```
jennybc, line 133: In this section, can you lean more heavily on the vocabulary we used for this subject in the column type guessing vignette?

Specifically, the notion of **edition** has to be made super clear and currently isn't mentioned here.

I don't think we need to actually link to that vignette, from here, once we get the wording right / complete.

sbearrows: I've added more content from the column type guessing vignette that I thought was relevant. Let me know what you think!

jennybc:
Here's an example of what I was talking about today: using some markup so that we don't advertise the installation of the dev version on the released site. I think readr is a good candidate for this approach.

https://github.com/tidyverse/glue/commit/7b0088531883e292b1e5c95667a1120c486f1d4e

sbearrows:
I think I've incorporated all your feedback so let me know how it looks when you have time

PR #1360, opened 2022-01-26
Use `rlang::is_interactive()`

### LINKED ISSUE #1356: Should `show_progress()` use `rlang::is_interactive()`?
https://github.com/tidyverse/readr/blob/a14f189a344f60113540d75f13b49432df12bb70/R/utils.R#L23

I naively tried `rlang::with_interactive(value = FALSE, {expression-that-calls-read_tsv()})` and was surprised that I still saw the indexing output from vroom. I guess this is because this sets the rlang specific interactive option, which readr/vroom won't recognize. 

This should also be thought about in tandem with `vroom_progress()`
https://github.com/r-lib/vroom/blob/eaab3e161c2d82b922217e7b19a315f7c939af18/R/vroom.R#L237

I know about the `progress` argument that I could set directly, but this still feels relevant

jennybc:
Closes #1356

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -18,7 +18,7 @@ is_syntactic <- function(x) make.names(x) == x
 #'   `options(readr.show_progress = FALSE)`.
 #' - The code is run in a non-interactive session, as determined by
 #'   [rlang::is_interactive()].
-#' - The code is run in an RStudio notebook chunk, as determined by the
+#' - The code is run in an RStudio notebook chunk, as determined by
 #'   `getOption("rstudio.notebook.executing")`.
 #' @export
 show_progress <- function() {
--- a/R/utils.R
+++ b/R/utils.R
@@ -23,7 +23,7 @@ is_syntactic <- function(x) make.names(x) == x
 #' @export
 show_progress <- function() {
   isTRUE(getOption("readr.show_progress")) &&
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -12,17 +12,23 @@ is_syntactic <- function(x) make.names(x) == x
...
-    interactive() && # an interactive session
-    !isTRUE(getOption("rstudio.notebook.executing")) && # Not running in an RStudio notebook chunk
-    !isTRUE(getOption("knitr.in.progress")) # Not actively knitting a document
+  isTRUE(getOption("readr.show_progress")) &&
+    is_interactive() &&
+    # some analysis re: rstudio.notebook.executing can be found in:
+    # https://github.com/r-lib/rlang/issues/1031
+    # TL;DR it's not consulted by is_interactive(), but probably should be
+    # consulted for progress reporting specifically
+    !isTRUE(getOption("rstudio.notebook.executing"))
```
jennybc, line 31: Apparently I analyzed this in the past and decided this option is probably relevant for deciding whether to show progress bars. But it's no good for deciding "general interactivity". So we keep this one.

DavisVaughan (approved):
Probably needs a vroom port too, but looks great!

PR #1365, opened 2022-02-03
switch argument order in std::string()

MichaelChirico:
I believe the fill constructor (6) is intended here:

https://www.cplusplus.com/reference/string/string/string/

### DIFF
--- a/src/Progress.h
+++ b/src/Progress.h
@@ -8,10 +8,6 @@
 
 inline int now() { return clock() / CLOCKS_PER_SEC; }
 
-inline std::string clearLine(int width = 50) {
-  return "\r" + std::string(width, ' ') + "\r";
-}
-
 inline std::string showTime(int x) {
   std::stringstream ss;
   if (x < 60) {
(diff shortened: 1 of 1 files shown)


DavisVaughan:
Is this function being used at all? 🤔

MichaelChirico:
not sure! our build tools identified the issue

jennybc:
Is there more context for this change, i.e. what motivates it in terms of visible behaviour?

MichaelChirico:
this was identified statically by a linter, not at runtime.

I'm not sure what the current code does but it seems clear from the doc that (into, char) works as intended. cc @hadley as original author from the blame.

hadley:
I have no recollection of this; it's quite likely I got it wrong and the code was never used. Might be easier to just delete it.

MichaelChirico:
@jennybc I'll follow the maintainer's lead, WDYT?

jennybc:
If it's dead code, then, yes let's just remove it.

MichaelChirico:
Thanks! removed the function

jennybc:
Thanks!

Issue #1372, opened 2022-02-28
Upkeep for readr


jennybc:
Pre-history

* [x] `usethis::use_readme_rmd()`
* [x] `usethis::use_roxygen_md()`
* [x] `usethis::use_github_links()`
* [ ] ~~`usethis::use_pkgdown_github_pages()`~~
* [x] `usethis::use_tidy_github_labels()`
* [x] `usethis::use_tidy_style()`
* [x] `usethis::use_tidy_description()`
* [x] `urlchecker::url_check()`

2020

* [x] `usethis::use_package_doc()`
Consider letting usethis manage your `@importFrom` directives here.
`usethis::use_import_from()` is handy for this.
* [x] `usethis::use_testthat(3)` and upgrade to 3e, [testthat 3e vignette](https://testthat.r-lib.org/articles/third-edition.html)
* [x] Align the names of `R/` files and `test/` files for workflow happiness.
`usethis::rename_files()` can be helpful.

2021

* [ ] ~~`usethis::use_tidy_dependencies()`~~
* [x] `usethis::use_tidy_github_actions()` and update artisanal actions to use `setup-r-dependencies`
* [x] Remove check environments section from `cran-comments.md`
* [x] Bump required R version in DESCRIPTION to 3.4
* [x] Use lifecycle instead of artisanal deprecation messages, as described in [Communicate lifecycle changes in your functions](https://lifecycle.r-lib.org/articles/communicate.html)

2022

* [x] `usethis::use_tidy_coc()`
* [ ] ~~Update pkgdown site using instructions at <https://tidytemplate.tidyverse.org>~~
* [ ] ~~Re-publish released site using [r-lib/pkgdown#2051](https://github.com/r-lib/pkgdown/pull/2051)~~
* [x] Ensure pkgdown `development` is `mode: auto` in pkgdown config
* [x] Handle and close any still-open `master` --> `main` issues
* [x] Update README badges, instructions in [r-lib/usethis#1594](https://github.com/r-lib/usethis/issues/1594)
* [ ] ~~Update errors to rlang 1.0.0~~ _opened a separate issue_ #1384

sbearrows:
Reposting so I can add checkmarks

* [ ] usethis::use_tidy_coc()
* [ ] Update errors to rlang 1.0.0
* [ ] Update pkgdown site using instructions at https://tidytemplate.tidyverse.org/
* [ ] Re-publish released site using https://github.com/r-lib/pkgdown/pull/2051
[...]

PR #1373, opened 2022-03-01
Update code of conduct

sbearrows:
Part of #1372


README.Rmd
```diff
@@ -178,4 +178,4 @@ Thanks to:
 
 ## Code of Conduct
 
-Please note that the readr project is released with a [Contributor Code of Conduct](https://readr.tidyverse.org/CONDUCT.html). By contributing to this project, you agree to abide by its terms.
```
jennybc, line 181: FWIW we've decided to no longer go out of our way to place this text in README, since GitHub features a code of conduct rather prominently, as do our pkgdown sites.

But we also aren't too worried about removing this text. It's a weak recommendation.

So we can remove this section. In which case you'll need to do `devtools::build_readme()` again.

jennybc (approved):

PR #1378, opened 2022-03-03
`use_tidy_style()`

sbearrows:
Part of #1372


R/col_types.R
```diff
@@ -308,16 +308,12 @@ colourise_cols <- function(cols, colourise = crayon::has_color()) {
     cols[[i]] <- switch(fname,
       col_skip = ,
       col_guess = cols[[i]],
-
```
jennybc, line 311: I think these blank lines are desirable. They seem to logically "chunk up" the different flavours of column (character-ish, logical, number-ish, datetime-ish). I'd keep this blank line and the ones right below (in the same call).

R/col_types.R
```diff
@@ -352,7 +348,6 @@ str.col_spec <- function(object, ..., indent.str = "") {
 
     # Append the current indentation string to the specs
     paste(indent.str, specs, collapse = "\n"),
-
```
jennybc, line 355: In contrast, there's no easily discernible motivation for this blank line, so I am willing to let it go.

R/write.R
```diff
@@ -336,10 +335,10 @@ format_delim <- function(x, delim, na = "NA", append = FALSE,
...
 #' @rdname format_delim
 format_csv <- function(x, na = "NA", append = FALSE, col_names = !append,
-  quote = c("needed", "all", "none"),
-  escape = c("double", "backslash", "none"),
-  eol = "\n",
-  quote_escape = deprecated()) {
+                       quote = c("needed", "all", "none"),
+                       escape = c("double", "backslash", "none"),
+                       eol = "\n",
+                       quote_escape = deprecated()) {
```
jennybc, line 341: I think we are actually changing our style here to match what was already present vs. what styler has done FYI.

That being said, I'm OK with going along with styler for now. I just want to mention this, so you won't be surprised in the future.

Applies to several similar diffs.

jennybc (commented):
There's just one place, in `R/col_types.R`, where I feel like we should decline styler's suggestions. There are some blank lines that enhance human readability, but styler doesn't know that.

jennybc (approved):
LGTM

PR #1379, opened 2022-03-03
Update package doc

sbearrows:
Part of #1372

This should be contained to changes around `usethis::use_package_doc()`


jennybc (changes requested):
There should still be a `man/readr-package.Rd` file, when all is said and done. I think you need to `document()`.

jennybc (approved):

PR #1380, opened 2022-03-03
`usethis::use_tidy_dependencies()`

sbearrows:
Part of #1372


jennybc:
I don't see any good reason to "get out ahead" of actual usage of the various packages and functions here. So I suggest we stand down and close this without merging.

But it's a good thought experiment as we look towards the future. When working on readr/vroom, we should have a very low threshold for taking on any of the dependencies that `use_tidy_dependencies()` would bring in. As we touch various surfaces of these packages, we should lean towards adopting the emerging tidyverse-wide solutions vs. continuing to maintain any DIY solution.

In particular, I think we should not hesitate to import rlang and start using its functions in an unqualified manner.

Issue #1389, opened 2022-03-17
Inconsistent `include_na` behaviour when factor levels are given


jennybc:
Given our recent work on factors, I was working on making the docs more clear and adding some examples.

And I found some puzzling behaviour. It's not a readr 1e vs readr 2e / vroom issue. I guess this has always been weird.

``` r
library(readr)

x <- c("a", "b", "NA")

# element 3 is NA because it matches an `na` string
# no warning, no problems, NA is a factor level
with_edition(1,
  parse_factor(x, levels = c("a", "b"), na = c("", "NA"), include_na = TRUE)
)
#> [1] a    b    <NA>
#> Levels: a b <NA>
with_edition(2,
  parse_factor(x, levels = c("a", "b"), na = c("", "NA"), include_na = TRUE)
)
#> [1] a    b    <NA>
#> Levels: a b <NA>

# element 3 is NA because its value is not found in the levels
# YES warning, YES problems, NA is NOT a factor level, despite include_na = TRUE
with_edition(1,
  parse_factor(x, levels = c("a", "b"), na = "", include_na = TRUE)
)
#> Warning: 1 parsing failure.
#> row col           expected actual
#>   3  -- value in level set     --
#> [1] a    b    <NA>
#> attr(,"problems")
#> # A tibble: 1 × 4
#>     row   col expected           actual
#>   <int> <int> <chr>              <chr> 
#> 1     3    NA value in level set NA    
#> Levels: a b
with_edition(2,
  parse_factor(x, levels = c("a", "b"), na = "", include_na = TRUE)
)
#> Warning: 1 parsing failure.
#> row col           expected actual
#>   3  -- value in level set     --
#> [1] a    b    <NA>
#> attr(,"problems")
#> # A tibble: 1 × 4
#>     row   col expected           actual
#>   <int> <int> <chr>              <chr> 
#> 1     3    NA value in level set NA    
#> Levels: a b
```

<sup>Created on 2022-03-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

My motivation was wondering if `include_na` had any effect when `levels` are given.

It would seem reasonable to ignore `include_na` in this case. The user **has** provided their desired levels. If they want `NA` in there, they should include `NA` in the levels.

However, that's not what `parse_factor()` does.

It seems to honour `include_na` for an "early `NA"`, i.e. a string that is replaced with `NA` due to matching one of the `na` strings.

[...]

PR #1390, opened 2022-03-17
Improve docs for `parse_factor()` / `col_factor()`

jennybc:
Recent factor work made me realize some of the docs around factor handling were confusing or ambiguous. This PR is meant to tighten things up.

vroom inherits the relevant parameter documentation from readr, so this basically covers both packages.

### DIFF
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -220,7 +220,7 @@ guess_parser <- function(x, locale = default_locale(), guess_integer = FALSE, na
 #'
 #' @param levels Character vector of the allowed levels. When `levels = NULL`
 #'   (the default), `levels` are discovered from the unique values of `x`, in
-#'   the order in which they are encountered in `x`.
+#'   the order in which they appear in `x`.
 #' @param ordered Is it an ordered factor?
 #' @param include_na If `TRUE` and `x` contains at least one `NA`, then `NA`
 #'   is included in the levels of the constructed factor.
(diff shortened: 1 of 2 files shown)


R/collectors.R
```diff
@@ -214,33 +214,41 @@ guess_parser <- function(x, locale = default_locale(), guess_integer = FALSE, na
...
-#' @param levels Character vector providing set of allowed levels. if `NULL`,
-#'   will generate levels based on the unique values of `x`, ordered by order
-#'   of appearance in `x`.
+#' @param levels Character vector of the allowed levels. When `levels = NULL`
+#'   (the default), `levels` are discovered from the unique values of `x`, in
+#'   the order in which they are encountered in `x`.
 #' @param ordered Is it an ordered factor?
-#' @param include_na If `NA` are present, include as an explicit factor to level?
+#' @param include_na If `TRUE` and `x` contains at least one `NA`, then `NA`
+#'   is included in the levels of the constructed factor.
```
jennybc, line 226: This is still not totally correct or complete, but I think it's an improvement. Once we sort out #1389, we can revisit this.

R/collectors.R
```diff
@@ -214,33 +214,41 @@ guess_parser <- function(x, locale = default_locale(), guess_integer = FALSE, na
...
+#' parse_factor(c("a", "b"), levels = letters[1:5])
 #'
-#' # Base R factor() silently converts unknown levels to NA
-#' x1 <- factor(x, levels)
+#' x <- c("cat", "dog", "caw")
+#' animals <- c("cat", "dog", "cow")
 #'
-#' # parse_factor generates a warning & problems
-#' x2 <- parse_factor(x, levels)
+#' # base::factor() silently converts elements that do match any levels to NA
```
sbearrows, line 246: I think this is a typo?

sbearrows (changes requested):
I think just that typo but otherwise definitely in better shape!

PR #1400, opened 2022-05-01
Correct the explanation of `should_read_lazy`

eitsupi:


jennybc:
Thanks for this PR. Once I reviewed it, it became clear those docs needed even more love, so I did that. Thanks for bringing my attention to it!

eitsupi:
@jennybc I just noticed that the following sections also need to be changed.
https://github.com/tidyverse/readr/blob/eb14f9523ef18a73f533fe6c4031fd5a95740b5a/R/read_delim.R#L85-L90

PR #1422, opened 2022-08-22
Update test for `NA` quoting

sbearrows:
When we made the changes necessary for https://github.com/tidyverse/vroom/pull/436 this caused test failures in readr that are outdated since we are no longer quoting `NA`s

Because `readr::format_csv()` uses `vroom::vroom_format()` the tests for https://github.com/tidyverse/vroom/pull/436 I think should stay there rather than also being in readr.

### DIFF
--- a/tests/testthat/test-write.R
+++ b/tests/testthat/test-write.R
@@ -7,9 +7,14 @@ test_that("strings are only quoted if needed", {
   expect_equal(ssv, "a\n,\n")
 })
 
+test_that("a literal NA is quoted", {
+  skip_if_edition_second()
+  expect_equal(format_csv(data.frame(x = "NA")), "x\n\"NA\"\n")
+})
+
 test_that("na argument modifies how missing values are written", {
-  df <- data.frame(x = c(NA, "x"), y = c(1, 2))
-  expect_equal(format_csv(df, na = "None"), "x,y\nNone,1\nx,2\n")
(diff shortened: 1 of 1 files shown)


tests/testthat/test-write.R
```diff
@@ -7,13 +7,9 @@ test_that("strings are only quoted if needed", {
   expect_equal(ssv, "a\n,\n")
 })
 
-test_that("a literal NA is quoted", {
-  expect_equal(format_csv(data.frame(x = "NA")), "x\n\"NA\"\n")
-})
-
```
jennybc, line 13: This test still applies to the first edition, though. Instead of removing it, I recommend inserting `skip_if_edition_second()`, which is used a lot elsewhere in this test suite.

jennybc (commented):
> Because `readr::format_csv()` uses `vroom::vroom_format()` ...

Just to make sure we're on the same page, `readr::format_csv()` calls `vroom::vroom_format()` by default these days, but not if someone has requested the first edition.

The proposed modification to the second test looks good, in that it makes the expected output not edition-specific. I think I'd go ahead and keep the `NA` in the numeric column, as that's presumably intended to test the behaviour in character and numeric columns.

tests/testthat/test-write.R
sbearrows: Yes you're right I had forgotten about that

sbearrows:
@jennybc Any other TODO's here?

jennybc (approved):

PR #1426, opened 2022-08-26
add examples of quiet `name_repair`

### LINKED ISSUE #1405: readr does not expose argument `quiet` in the calls to vec_as_names
See https://github.com/r-lib/vctrs/issues/1547#issuecomment-1149851253

Issue inherited from `vroom`, which calls `vec_as_names` in a way that can't be configured.

https://github.com/tidyverse/vroom/blob/8bdd2b5dbfa7d4c4082226ba7b81ca54fded4e0f/R/col_types.R#L389

Software that calls readr many times can face very verbose messaging that is cumbersome to suppress.

sbearrows:
Fixes #1405

We describe using functions for custom name repair in documentation but don't include any examples of doing so. These examples show how to quiet `vctrs::vec_as_names()` without exposing `quiet=` to the user which doesn't feel worth it since there is a method to quiet name repair already. This is also a relatively rare use case.

### DIFF
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -139,6 +139,20 @@ NULL
 #' y
 #' problems(y)
 #'
+#' # Column names --------------------------------------------------------------
+#' # By default, readr duplicate name repair is noisy
+#' read_csv(I("x,x\n1,2\n3,4"))
+#'
+#' # To quiet, use a function that does name repair quietly
+#' quiet_repair <- function(x){
+#'   vctrs::vec_as_names(x, repair = "unique", quiet = TRUE)
+#' }
+#'
+#' read_csv(I("x,x\n1,2\n3,4"), name_repair = quiet_repair)
+#'
+#' # Or use "minimal" to turn off name repair
+#' read_csv(I("x,x\n1,2\n3,4"), name_repair = "minimal")
+#'
 #' # File types ----------------------------------------------------------------
 #' read_csv(I("a,b\n1.0,2.0"))
 #' read_csv2(I("a;b\n1,0;2,0"))
--- a/man/read_delim.Rd
+++ b/man/read_delim.Rd
@@ -302,6 +302,20 @@ y <- read_csv(I("x\n1\n2\nb"), col_types = list(col_double()))
 y
 problems(y)
 
+# Column names --------------------------------------------------------------
+# By default, readr duplicate name repair is noisy
+read_csv(I("x,x\n1,2\n3,4"))
+
+# To quiet, use a function that does name repair quietly
+quiet_repair <- function(x){
+  vctrs::vec_as_names(x, repair = "unique", quiet = TRUE)
+}
+
+read_csv(I("x,x\n1,2\n3,4"), name_repair = quiet_repair)
+
+# Or use "minimal" to turn off name repair
+read_csv(I("x,x\n1,2\n3,4"), name_repair = "minimal")
+
 # File types ----------------------------------------------------------------
 read_csv(I("a,b\n1.0,2.0"))
 read_csv2(I("a;b\n1,0;2,0"))


DavisVaughan:
@sbearrows you are getting errors in the build because vctrs isn't in Imports/Suggests for readr (but it is for vroom, which is where `vctrs::vec_as_names()` actually gets called). I think vctrs could be added to Suggests for readr to fix that

DESCRIPTION
```diff
@@ -49,7 +49,8 @@ Suggests:
     tzdb (>= 0.1.1),
     waldo,
     withr,
-    xml2
+    xml2,
+    vctrs
```
jennybc, line 53: Nitpick: our current style is to always keep packages in DESCRIPTION in alphabetical order. So vctrs should be inserted in the appropriate place in the list.

Description of this style and why:
https://r-pkgs.org/dependencies.html#run-time-vs-develop-time-dependencies

Note also the advice there re: usethis functions that help keep DESCRIPTION in line with these conventions.

jennybc (approved):
Reading up on this gives me some thoughts about name repair, which I'll open another issue about, probably in vctrs. But because of that, can you change the wording here from "fixes" to "relates to", so that merging this PR does **not** close #1405?

But, yes, documenting how a user can do this is certainly an immediate improvement we can make.

It feels like equivalent examples should also appear in vroom. The usual problem of having readr expose vroom features.

jennybc (changes requested):
I've returned to this, in order to open that other issue/discussion mentioned above.

And reading up on `vctrs::vec_as_names()` reminds me that there is an option one can set.

In terms of documenting how someone can silence name repair messaging, I'm not convinced that creating a custom function (and having to show and Suggest vctrs) is superior to just pointing out that there is an option that controls this.

``` r
library(readr)

# let's suppress this noise for the moment, to focus on name repair 
withr::local_options(list(readr.show_col_types = FALSE))

read_csv(I("x,x\n1,2\n3,4"))
#> New names:
#> • `x` -> `x...1`
#> • `x` -> `x...2`
#> # A tibble: 2 × 2
#>   x...1 x...2
#>   <dbl> <dbl>
#> 1     1     2
#> 2     3     4

withr::with_options(list(rlib_name_repair_verbosity = "quiet"),
  read_csv(I("x,x\n1,2\n3,4"))
)
#> # A tibble: 2 × 2
#>   x...1 x...2
#>   <dbl> <dbl>
#> 1     1     2
#> 2     3     4

# for folks who don't use withr
op <- options(rlib_name_repair_verbosity = "quiet")
read_csv(I("x,x\n1,2\n3,4"))
#> # A tibble: 2 × 2
#>   x...1 x...2
#>   <dbl> <dbl>
#> 1     1     2
#> 2     3     4
options(op)
```

<sup>Created on 2022-08-29 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

So I'm un-approving, because now I think I prefer adding the above as an example. It's a weak preference, though. Or we could show both?

(tagging @DavisVaughan explicitly, because I'm not sure if he's subscribed to this conversation)

DavisVaughan:
Silencing through the global option and `withr/rlang::with_options()` seems fine to me as a less complicated alternative to creating a custom repair function.

I wouldn't show the `options()` case, as I don't want to encourage turning this option on globally (i.e. it is nice to keep the quiet name repair scoped down to the smallest amount of code you'd like to silence)

sbearrows:
Alright this looks more straightforward, so I'll go ahead and make those changes

jennybc:
> I wouldn't show the options() case, as I don't want to encourage turning this option on globally (i.e. it is nice to keep the quiet name repair scoped down to the smallest amount of code you'd like to silence)

Yeah, I agree. And withr is already in readr's Suggests and is already indirectly a hard dependency via vroom.

DESCRIPTION
jennybc: This is now irrelevant in this PR, but this convention should be noted for future changes to dependencies.

jennybc:
BTW I made a proposal in vctrs which would be relevant to readr and vroom, if it goes anywhere: https://github.com/r-lib/vctrs/issues/1629

jennybc (approved):

PR #1431, opened 2022-09-02
Rework `problems` example in "Getting Started"

### LINKED ISSUE #1398: readr no longer reproduces the problems from challenge.csv
`challenge.csv` is designed to teach some key challenges of parsing and features of `readr`. However, the example is broken.

```r
challenge <- read_csv(readr_example("challenge.csv"))
#> 
#> ── Column specification ────────────────────────────────────────────────────────
#> cols(
#>   x = col_double(),
#>   y = col_logical()
#> )
#> Warning: 1000 parsing failures.
#>  row col           expected     actual                                                           file
#> 1001   y 1/0/T/F/TRUE/FALSE 2015-01-16 '/Users/runner/work/_temp/Library/readr/extdata/challenge.csv'
#> 1002   y 1/0/T/F/TRUE/FALSE 2018-05-18 '/Users/runner/work/_temp/Library/readr/extdata/challenge.csv'
#> 1003   y 1/0/T/F/TRUE/FALSE 2015-09-05 '/Users/runner/work/_temp/Library/readr/extdata/challenge.csv'
#> 1004   y 1/0/T/F/TRUE/FALSE 2012-11-28 '/Users/runner/work/_temp/Library/readr/extdata/challenge.csv'
#> 1005   y 1/0/T/F/TRUE/FALSE 2020-01-13 '/Users/runner/work/_temp/Library/readr/extdata/challenge.csv'
#> .... ... .................. .......... ..............................................................
#> See problems(...) for more details.
```
[source](https://r4ds.had.co.nz/data-import.html)

Here's what I get when I run this code on my computer (this is a screen cap from the [vignette](https://cran.r-project.org/web/packages/readr/vignettes/readr.html), but I have the same issue on my computer)
Screen Shot 2022-04-17 at 9 46 39 PM

sbearrows:
Fixes #1398 

Modifies the `challenge.csv` file so that it actually reports some parsing problems. Also, updates some of the wording around `spec_csv()` and `guess_max`

This PR doesn't aim to fix all of the issues with the "Getting Started" page and in fact we already have a PR open for that https://github.com/tidyverse/readr/pull/1364. It only deals with making it better than before and fixing things that directly relate to `challenge.csv`.


vignettes/readr.Rmd
```diff
@@ -142,15 +142,6 @@ mtcars_spec
 cols_condense(mtcars_spec)
 ```
 
-
-By default readr only looks at the first 1000 rows. This keeps file parsing speedy, but can generate incorrect guesses. For example, in `challenge.csv` the column types change in row 1001, so readr guesses the wrong types. One way to resolve the problem is to increase the number of rows:
-
```
sbearrows, line 147: This is no longer true so it's been removed. Now, readr (via vroom) selects semi-randomized rows to select for guess, and always includes the last row for guessing. So increasing `guess_max` won't necessarily resolve problems in this case.

https://github.com/tidyverse/vroom/blob/548344f4826818ca988e80b90cdf7fd837a90561/src/collectors.h#L161-L189

vignettes/readr.Rmd
```diff
@@ -160,60 +151,38 @@ readr comes with five parsers for rectangular file formats:
 * `read_fwf()` for fixed-width files
 * `read_log()` for web log files
 
-Each of these functions firsts calls `spec_xxx()` (as described above), and then parses the file according to that column specification:
```
sbearrows, line 163: No longer true, `spec_csv()` is a first edition readr function.

vignettes/readr.Rmd
```diff
@@ -160,60 +151,38 @@ readr comes with five parsers for rectangular file formats:
...
 
 ```{r}
-df1 <- read_csv(readr_example("challenge.csv"))
+df1 <- read_csv(readr_example("challenge.csv"), col_types = list(
+  date = col_date(),
+  value = col_double()
+))
 ```
 
-The rectangular parsing functions almost always succeed; they'll only fail if the format is severely messed up. Instead, readr will generate a data frame of problems. The first few will be printed out, and you can access them all with `problems()`:
```
sbearrows, line 169: Updated text to reflect current behavior for `problems()` which does not print the first few parsing problems out.

vignettes/readr.Rmd
```diff
@@ -160,60 +151,38 @@ readr comes with five parsers for rectangular file formats:
...
-#>   y = col_character()
-#> )
-```
-
-You can also access it after the fact using `spec()`:
-
-```{r}
-spec(df1)
-spec(df2)
+In this example, one of the dates is in an incorrect format.
```
sbearrows, line 172: This could use some garnishing

vignettes/readr.Rmd
```diff
@@ -160,60 +151,38 @@ readr comes with five parsers for rectangular file formats:
...
+Since `problems()` gives us the row number from the original file, fixing this value in the original file could be easy, especially since there is only one problematic value.
+But data in the wild is rarely so clean.
+We can fix this parsing error by importing the date column as character and using the lubridate package to parse our dates.
+
+```{r, eval = FALSE}
+df1 <- read_csv(readr_example("challenge.csv"), col_types = list(
+  date = col_character(),
+  value = col_double()
+))
+df1$date <- lubridate::ymd(df1$date)
```
sbearrows, line 182: Even though we can't run this due to the lubridate call, I think it serves as a good example of a problem you'd see in the wild (ie mixed bag of date formats). And here is an example of such a problem:
https://github.com/tidyverse/readr/issues/737

vignettes/readr.Rmd
```diff
@@ -160,60 +151,38 @@ readr comes with five parsers for rectangular file formats:
...
-Each of these functions firsts calls `spec_xxx()` (as described above), and then parses the file according to that column specification:
+If readr finds problems with parsing your data, it will emit a warning message.
+For example, in our `challenge.csv` file, the column type we expected from our data was a column of dates and a column of doubles.
 
 ```{r}
-df1 <- read_csv(readr_example("challenge.csv"))
+df1 <- read_csv(readr_example("challenge.csv"), col_types = list(
+  date = col_date(),
+  value = col_double()
+))
```
sbearrows, line 161: Note that this would still generate parsing errors even if we don't include/force `col_types` so we could just have the following if we think it makes for a better narrative:
```r
df1 <- read_csv(readr_example("challenge.csv"))
```
But I think it does drive the point across that supplying a `col_spec` is safer.

vignettes/readr.Rmd
```diff
@@ -160,60 +151,38 @@ readr comes with five parsers for rectangular file formats:
...
+df1 <- read_csv(readr_example("challenge.csv"), col_types = list(
+  date = col_character(),
+  value = col_double()
+))
+df1$date <- lubridate::ymd(df1$date)
 ```
 
-(This also allows you to access the full column specification if you're reading a very wide file. By default, readr will only print the specification of the first 20 columns.)
-
-If you want to manually specify the column types, you can start by copying and pasting this code, and then tweaking it fix the parsing problems.
```
sbearrows, line 204: We could bring this part back over if you think it's important/useful.

vignettes/readr.Rmd
```diff
@@ -160,60 +151,38 @@ readr comes with five parsers for rectangular file formats:
...
-
-```{r}
-df2 <- read_csv(readr_example("challenge.csv"), guess_max = 1001)
-```
-
-Another approach is to manually supply the column specification.
-
-### Overriding the defaults
-
-In the previous examples, you may have noticed that readr printed the column specification that it used to parse the file:
```
sbearrows, line 185: This is no longer true; readr does not print the column specification in a form that is copy/pastable into `col_types`.

sbearrows:
Notes:

- No `NEWS` bullet was added because AFAIR we don't add `NEWS` bullets for updating docs
- I believe everything I have touched and made changes to, is consistent with behaviors in vroom. That doesn't mean the entire document is consistent with vroom behaviors though.
- There are a few other documentation issues we have coming up that we could tack on to this PR if we decide they should go in "Get Started": https://github.com/tidyverse/readr/issues/1369, https://github.com/tidyverse/readr/issues/1424, https://github.com/tidyverse/readr/issues/1377

vignettes/readr.Rmd
jennybc: Yeah I'm fine with showing the use of lubridate. I think you should add some brief comment that we can't execute this chunk, since readr doesn't depend on lubridate. But that it's indicative of a general type of workaround, where you import as character and use a more specialized package (or custom function) to parse a tricky column.

vignettes/readr.Rmd
jennybc: I actually think the idea of copy/pasting a spec does still belong here, especially since we're already advising folks to provide an explicit col spec.

vignettes/readr.Rmd
```diff
@@ -253,12 +222,12 @@ Use the `col_types` argument to override the default choices. There are two ways
     
     ```r
       read_csv("iris.csv", col_types = list(
-      Sepal.Length = "d",
-      Sepal.Width = "d",
-      Petal.Length = "d",
-      Petal.Width = "d",
-      Species = col_factor(c("setosa", "versicolor", "virginica"))
-    ))
+        Sepal.Length = "d",
```
jennybc, line 225: Let's convert this to a normal chunk, with `eval = FALSE`

In general, this is good for keeping formatting and styling consistent.

jennybc (commented):
I made some comments inline. Overall, it's looking good: removing lots of incorrect content and updating it.

Zooming out:

Can you describe what changes you made to challenge.csv? And how does that change what we see in the rendered vignette? I can't tell that from reading this PR, i.e. I would have to fetch the PR and render the vignette.

One other concern: changing a dataset is potentially a breaking change. I.e. if someone, somewhere, uses the existing challenge.csv dataset, this will break their code. Did you do any research into this? For example, searching CRAN packages for usage of this dataset and just googling for "challenge.csv".
 
I could imagine this is basically an internal matter and it's OK to change, but we should do some due diligence. It could also show up when we run revdeps, but it's nicer to learn this sooner rather than later.

sbearrows:
@jennybc The file has a column of dates and a column of doubles. One of the dates is formatted as `12032020`. All the other dates in the file are of the form `12-03-2020`. I had to select a specific line where `guess_max` wouldn't look to get the parsing error.

I've made changes to address most of your comments but I still need to look into whether other packages are using `challenge.csv`.

sbearrows:
@jennybc I couldn't find any other packages that use `challenge.csv`. The only reference to it I could find is: https://search.r-project.org/CRAN/refmans/csvwr/html/datatype_to_type.html where they use it in an example but it's set to `dontrun`

vignettes/readr.Rmd
```diff
@@ -127,7 +127,7 @@ parse_number("$1,234")
 
 There are two parsers that will never be guessed: `col_skip()` and `col_factor()`. You will always need to supply these explicitly.
 
-You can see the specification that readr would generate for a column file by using `spec_csv()`, `spec_tsv()` and so on:
+You can see the specification that readr would generate for a file by using `spec_csv()`, `spec_tsv()`, and so on.
 
 ```{r}
 x <- spec_csv(readr_example("challenge.csv"))
```
jennybc, line 133: Wouldn't it make sense to reveal the spec? I.e. why are we assigning the result to `x`?

(I have fetched the PR and rendered the vignette now. I assume you are also looking at the rendered result as you go.)

vignettes/readr.Rmd
```diff
@@ -127,7 +127,7 @@ parse_number("$1,234")
 
 There are two parsers that will never be guessed: `col_skip()` and `col_factor()`. You will always need to supply these explicitly.
```
jennybc, line 128: Isn't it also true the `col_integer()` will not be guessed? I spotted this since we are working here in the immediate neighborhood.

vignettes/readr.Rmd
jennybc: The rows are not selected in a random fashion. They're basically evenly spaced within the file and the last row is always included for guessing.

https://github.com/tidyverse/vroom/blob/0e2df4cc8a07fbb007206f6bfcc6a3732e6c4613/src/collectors.h#L164

vignettes/readr.Rmd
```diff
@@ -142,14 +142,13 @@ mtcars_spec
 cols_condense(mtcars_spec)
 ```
 
-
-By default readr only looks at the first 1000 rows. This keeps file parsing speedy, but can generate incorrect guesses. For example, in `challenge.csv` the column types change in row 1001, so readr guesses the wrong types. One way to resolve the problem is to increase the number of rows:
+When guessing column types, readr selects rows to use for guessing semi-randomly. The default number of rows it selects is 1000. If the column type guess is incorrect, increasing `guess_max` may or may not improve the results.
 
 ```{r}
 x <- spec_csv(readr_example("challenge.csv"), guess_max = 1001)
```
jennybc, line 148: Similar to above, if we're trying to make a point about the spec, we should reveal it.

vignettes/readr.Rmd
```diff
@@ -160,61 +159,46 @@ readr comes with five parsers for rectangular file formats:
...
+For example, in our `challenge.csv` file, the column type we expected from our data was a column of dates and a column of doubles.
 
 ```{r}
 df1 <- read_csv(readr_example("challenge.csv"))
 ```
 
-The rectangular parsing functions almost always succeed; they'll only fail if the format is severely messed up. Instead, readr will generate a data frame of problems. The first few will be printed out, and you can access them all with `problems()`:
+The data was successfully imported, but we get a warning message about parsing issues.
+To view more details about these parsing issues, call `problems()` on your data frame.
+This will generate a data frame with information about where the problem was found, and why.
```
jennybc, line 171:
```suggestion
This will generate a data frame with information about the parsing problems, with one row per problem.
```

The previous proposed wording seemed too specific to this example, i.e. that there would be exactly 1 problem.

vignettes/readr.Rmd
```diff
@@ -160,61 +159,46 @@ readr comes with five parsers for rectangular file formats:
...
+In this example, one of the dates is in an incorrect format.
+Since `problems()` gives us the row number from the original file, we can view that data in R with a combination of `readr::read_lines()` and `writeLines()`.
 
 ```{r}
-#> Parsed with column specification:
-#> cols(
-#>   x = col_integer(),
-#>   y = col_character()
-#> )
+writeLines(read_lines(readr_example("challenge.csv"), skip = 1325, n_max = 3))
```
jennybc, line 181: 1325 feels a bit like a Magic Number here. It feels like we should probably get it programmatically, since we do have access to it and explain that we're inspecting the neighborhood right around the problem line.

vignettes/readr.Rmd
```diff
@@ -160,61 +159,46 @@ readr comes with five parsers for rectangular file formats:
...
 ```
 
-You can also access it after the fact using `spec()`:
+We can work around this parsing error by importing the date column as character and converting it to a column of dates, after the fact. The lubridate package is a great option that we show here. But since readr doesn't depend on lubridate, we won't execute this code chunk.
```
jennybc, line 184: You actually could make this a static code chunk. That is, set it to `eval = FALSE`, but inline the current result. It's not a perfect solution, but the pros would seem to outweigh the cons.

I'm talking about showing folks this:

``` r
> df1[1325:1327, ]
# A tibble: 3 × 2
  date       value
  <date>     <dbl>
1 2008-03-16   762
2 2008-02-17   206
3 2019-01-28  1333
```

jennybc: I reiterate this:

> I think you should add some brief comment ... that it's indicative of a general type of workaround, where you import as character and use a more specialized package (or custom function) to parse a tricky column.

vignettes/readr.Rmd
```diff
@@ -239,32 +223,32 @@ Use the `col_types` argument to override the default choices. There are two ways
...
-    ```
+```{r, eval = FALSE}
+read_csv("iris.csv", col_types = list(
+  Sepal.Length = col_double(),
+  Sepal.Width = col_double(),
+  Petal.Length = col_double(),
+  Petal.Width = col_double(),
+  Species = col_factor(c("setosa", "versicolor", "virginica"))
+))
+```
```
jennybc, line 234: This has messed up the intended indenting. Here and below.

jennybc (commented):
I left some inline comments.

--- 

High level comment about the new challenge.csv.

Does challenge.csv need to be 2000 lines long anymore?

Originally, I guess it needed to be over 1000 lines long to make a point about column type guessing for the `y` column, which was missing for the first 1000 lines and then started to have dates in it. (That still doesn't fully explain why the file was 2000 lines long, but I guess that will remain a mystery.)

What is the purpose of the `value` column in the new challenge.csv?

I notice that the original `x` column was integers only for the first 1000 lines and then started to have doubles in it. So I think it must be or have been used to demonstrate something about integers vs doubles.

But I can't figure out the purpose of `value` in the proposed challenge.csv, which doesn't seem to have this mix of integers and doubles.

---

Our modern approach to creating datasets in packages is to save the code that generates the dataset below `data-raw/`. It helps with exercises like this one, i.e. it shows how to regenerate a dataset and (hopefully) leave some breadcrumbs re: why the data is the way it is.

This is explained in the Data chapter of R Packages and, as also described there, supported by functions like `use_data_raw()`.

vignettes/readr.Rmd
sbearrows: I know, but I felt like being overly specific would be more confusing. I can modify to say "interspersed through-out the file and always includes the first and last row" instead.

sbearrows: Actually, this isn't technically true since `guess_max` doesn't guess column types using rows it will never read and once we complete https://github.com/tidyverse/readr/pull/1436 then `n_max = guess_max`. So if `guess_max = 1500` then `n_max = 1500` and it will guess using the first 1500 rows in the data. So `spec_*()` guessing will never be interspersed throughout the file, it'll always use all of the data it is given to guess. Here is an example using the original `challenge.csv` where the column `y` is filled with `NA` until row 1001.

``` r
# after we complete #1436
spec_csv(readr_example("challenge.csv"), guess_max = 1000) # n_max = 1000
#> cols(
#>   x = col_double(),
#>   y = col_logical()
#> )
# increasing guess_max does change the spec
spec_csv(readr_example("challenge.csv"), guess_max = 1001) # n_max = 1001
#> cols(
#>   x = col_double(),
#>   y = col_date(format = "")
#> )

# compared to how guessing works in read_csv
# since it's dispersed throughout the file, it guesses correctly
spec(read_csv(readr_example("challenge.csv"), guess_max = 1000, show_col_types = FALSE)) # n_max = all the data
#> cols(
#>   x = col_double(),
#>   y = col_date(format = "")
#> )
```

vignettes/readr.Rmd
sbearrows: This is how the example was when I started editing the file and I didn't see a reason to modify it since we aren't doing a total overhaul right now and the goal is to ensure that the `problems()` example has actual problems. Also, the behavior of `spec_*()` is currently in-flight https://github.com/tidyverse/readr/pull/1436 so some of the content here might change.

vignettes/readr.Rmd
sbearrows: No, I don't think so. Here is where the guessing happens https://github.com/tidyverse/vroom/blob/0c2423eb3bba77b05ef9fda1830f1f798cc2fedd/src/guess_type.cc#L128-L146

sbearrows:
> Does challenge.csv need to be 2000 lines long anymore?

I created the new challenge.csv by modify pieces of the old challenge.csv so that's one reason why it's also 2000 lines long. It also made it easier to insert a parsing error into the file and makes the data feel more realistic.


> What is the purpose of the value column in the new challenge.csv?...I can't figure out the purpose of value in the proposed challenge.csv, which doesn't seem to have this mix of integers and doubles.

The value column isn't being used to teach anything, its purpose is to make the data feel real since data with only one column would be odd. We could make this column also teach some other lesson about `problems()` or parsing if that's what you're suggesting. But I think we should only have one type of parsing problem for the example otherwise it might feel overloaded with information.



> Our modern approach to creating datasets in packages is to save the code that generates the dataset below data-raw/

I'll make a note to complete this TODO once we've landed on a dataset

vignettes/readr.Rmd
jennybc: The statement should not be technically misleading (e.g. suggesting that it's random and might select different rows each time you call it), but otherwise it's fine to stay high-level. So, yes, something along these lines: "interspersed through-out the file and always includes the first and last row".

Recall that we did create an entire new vignette on column type guessing, in case it makes sense to link to that (or borrow test from that), for more details: https://readr.tidyverse.org/articles/column-types.html.

To zoom out, I think a guiding principle is to avoid making this text so specific that it's tightly coupled with detail work like #1436. What level of understanding and detail is useful for the typical user?

The observation above about `n_max` and `guess_max` in `spec()` brings up an important question to think about over in #1436. Should `spec_*()` produce the same column type guesses as `read_*()` at default settings?

jennybc:
Given that the only column we're currently using for a concrete demo is a date column, I think we should just leave the legacy challenge.csv alone and create a new example file with exactly what you need. 

As I understand it, you want a single date column with at least one strategically positioned entry in a nonstandard format. This tricky entry should not cause the column to be guessed as character (so that's dictates its position), but it should create a parsing problem. Make it minimal and sufficient to support what you want to show. Store the `.R` script used to generate it below `data-raw/`.

Issue #1415, opened 2022-08-05
`parse_number` returns first number block only


tsufz:
Please briefly describe your problem and what output you expect. If you have a question, please don't use this form. Instead, ask on <https://stackoverflow.com/> or <https://community.rstudio.com/>.

Hi, I observe a problem with `parse_number`. If I type `parse_number("137-12-3")`, I expect the result `137123`. However, I get the result `137`. Obviously, only the first block, the rest is truncated. I checked this with the following examples:
```
parse_number("137-12-3") = 137
parse_number("t15t15") = 15
```
--
R version 4.2.1 (2022-06-23 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19044)

Matrix products: default

locale:
[1] LC_COLLATE=English_United Kingdom.utf8 
[2] LC_CTYPE=English_United Kingdom.utf8   
[3] LC_MONETARY=English_United Kingdom.utf8
[4] LC_NUMERIC=C                           
[5] LC_TIME=English_United Kingdom.utf8    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] forcats_0.5.1   stringr_1.4.0   dplyr_1.0.9     purrr_0.3.4    
[5] readr_2.1.2     tidyr_1.2.0     tibble_3.1.8    ggplot2_3.3.6  
[9] tidyverse_1.3.2

loaded via a namespace (and not attached):
 [1] pillar_1.8.0        compiler_4.2.1      cellranger_1.1.0   
 [4] dbplyr_2.2.1        tools_4.2.1         lubridate_1.8.0    
 [7] jsonlite_1.8.0      googledrive_2.0.0   lifecycle_1.0.1    
[10] gargle_1.2.0        gtable_0.3.0        pkgconfig_2.0.3    
[13] rlang_1.0.4         reprex_2.0.1        DBI_1.1.3          
[16] cli_3.3.0           haven_2.5.0         xml2_1.3.3         
[19] withr_2.5.0         httr_1.4.3          generics_0.1.3     
[22] vctrs_0.4.1         fs_1.5.2            hms_1.1.1          
[25] googlesheets4_1.0.0 grid_4.2.1          tidyselect_1.1.2   
[28] glue_1.6.2          R6_2.5.1            fansi_1.0.3        
[31] readxl_1.4.0        modelr_0.1.8        tzdb_0.3.0         
[...]

PR #1432, opened 2022-09-06
Clarify docs for parse number

### LINKED ISSUE #1424: Clarify docs for `parse_number()` behavior with non-numeric characters
Developed from https://github.com/tidyverse/readr/issues/1415

We might want to clarify the docs for how `parse_number()` handles special characters found after the first valid number. Right now the docs for `parse_number()` says:

> This drops any non-numeric characters before or after the first number.

I think this is a bit confusing since there could be valid numbers after the first number, for example:

``` r
readr::parse_number("132aa000")
#> [1] 132
```

<sup>Created on 2022-08-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>
Maybe something like:

"This parses the first number it finds, dropping any non-numeric characters before the first number and all characters after the first number."

sbearrows:
Fixes #1424

Developed from #1415

Tightening up the docs for how `parse_number()` would parse a string that contains valid numbers and non-numeric characters. Specifically, to detail this situation, where `parse_number()` only parses the first number it finds:

```r
readr::parse_number("132aa000")
#> [1] 132
```

### DIFF
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -142,6 +142,7 @@ col_skip <- function() {
 #' ## These all return 1000
 #' parse_number("$1,000") ## leading $ and grouping character , ignored
 #' parse_number("euro1,000") ## leading non-numeric euro ignored
+#' parse_number("t1000t1000") ## only parses first number found
 #'
 #' parse_number("1,234.56")
 #' ## explicit locale specifying European grouping and decimal marks
--- a/R/collectors.R
+++ b/R/collectors.R
@@ -140,7 +140,7 @@ col_skip <- function() {
 #' @export
 #' @examples
 #' ## These all return 1000
-#' parse_number("$1,000") ## leading $ and grouping character ignored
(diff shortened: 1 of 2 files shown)


R/collectors.R
```diff
@@ -139,8 +140,9 @@ col_skip <- function() {
 #' @export
 #' @examples
 #' ## These all return 1000
-#' parse_number("$1,000") ## leading $ and grouping character , ignored
+#' parse_number("$1,000") ## leading $ and grouping character ignored
 #' parse_number("euro1,000") ## leading non-numeric euro ignored
+#' parse_number("t1000t1000") ## only parses first number found
```
sbearrows, line 145: This example was pulled from the original issue https://github.com/tidyverse/readr/issues/1415

sbearrows:
Nothing TODO in vroom since `parse_*()` are readr only functions

jennybc:
LGTM

PR #1434, opened 2022-09-08
Add examples for `col_select` and remove mention of `list()`

### LINKED ISSUE #1421: Remove all mentions of `list()` in docs and examples for `col_select`
Should add some examples of using `col_select` as well.

Also consider if 'readr::read_delim()' should inherit 'col_select' docs from vroom, since readr only supports `col_select` when it delegates to vroom.

sbearrows:
Fixes #1421

Add examples for `col_select` in readr since we currently don't have any.

### DIFF
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -10,6 +10,7 @@ NULL
 #' decimal point. This format is common in some European countries.
 #' @inheritParams datasource
 #' @inheritParams tokenizer_delim
+#' @inheritParams vroom::vroom
 #' @param col_names Either `TRUE`, `FALSE` or a character vector
 #'   of column names.
 #'
@@ -54,12 +55,6 @@ NULL
 #'    By default, reading a file without a column specification will print a
 #'    message showing what `readr` guessed they were. To remove this message,
 #'    set `show_col_types = FALSE` or set `options(readr.show_col_types = FALSE).
-#' @param col_select Columns to include in the results. You can use the same
-#'   mini-language as `dplyr::select()` to refer to the columns by name. Use
-#'   `c()` or `list()` to use more than one selection expression. Although this
-#'   usage is less common, `col_select` also accepts a numeric column index. See
-#'   [`?tidyselect::language`][tidyselect::language] for full details on the
-#'   selection language.
 #' @param id The name of a column in which to store the file path. This is
 #'   useful when reading multiple input files and there is data in the file
 #'   paths, such as the data collection date. If `NULL` (the default) no extra
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -120,6 +120,23 @@ NULL
 #' # Or directly from a string with `I()`
 #' read_csv(I("x,y\n1,2\n3,4"))
 #'
+#' # Column selection-----------------------------------------------------------
+#' # Pass column names or indexes directly to select them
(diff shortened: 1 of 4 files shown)


R/read_delim.R
```diff
@@ -125,6 +120,23 @@ NULL
...
+#'   readr_example("chickens.csv"),
+#'   col_select = c(starts_with("c"), last_col())
+#' )
+#'
+#' # You can also rename specific columns
+#' read_csv(
+#'   readr_example("chickens.csv"),
+#'   col_select = c(egg_yield = eggs_laid, everything())
+#' )
+#'
```
sbearrows, line 139: This is adapted from vrooms examples of using `col_select`.

vroom uses the `mtcars.csv` dataset but I noticed that readr has a different version of `mtcars.csv` that didn't work well showcasing `col_select` so I'm using the `chickens.csv` data

R/read_delim.R
```diff
@@ -54,12 +55,6 @@ NULL
 #'    By default, reading a file without a column specification will print a
 #'    message showing what `readr` guessed they were. To remove this message,
 #'    set `show_col_types = FALSE` or set `options(readr.show_col_types = FALSE).
-#' @param col_select Columns to include in the results. You can use the same
-#'   mini-language as `dplyr::select()` to refer to the columns by name. Use
-#'   `c()` or `list()` to use more than one selection expression. Although this
-#'   usage is less common, `col_select` also accepts a numeric column index. See
-#'   [`?tidyselect::language`][tidyselect::language] for full details on the
-#'   selection language.
```
sbearrows, line 62: Since we are inheriting docs from vroom now, this gets re-rendered the same, just without the mention of `list()`

sbearrows:
@jennybc So far, I've updated docs for `col_select` by inheriting them from vroom and added examples for `col_select`. I was thinking we could look over what I have now, make changes, and then I could look at the other arguments that should inherit docs from vroom and make those changes in this PR. 

Am I correct in thinking that our general strategy for arguments like `skip_empty_rows` that exist in readr ed 1 and ed 2, is to keep the docs in readr and not inherit from vroom. But, if the argument only exists in vroom, like `col_select`, those docs should inherit from vroom?

jennybc:
> and then I could look at the other arguments that should inherit docs from vroom and make those changes in this PR

Let's make sure that any other arguments that need similar treatment are identified and open an issue for that. But let's merge this as the solution for `col_select`.

> Am I correct in thinking that our general strategy for arguments like skip_empty_rows that exist in readr ed 1 and ed 2, is to keep the docs in readr and not inherit from vroom. But, if the argument only exists in vroom, like col_select, those docs should inherit from vroom?

Yeah that is certainly the way I'm currently thinking about it.

jennybc (approved):
> and then I could look at the other arguments that should inherit docs from vroom and make those changes in this PR

Let's make sure that any other arguments that need similar treatment are identified and open an issue for that. But let's merge this as the solution for `col_select`.

> Am I correct in thinking that our general strategy for arguments like skip_empty_rows that exist in readr ed 1 and ed 2, is to keep the docs in readr and not inherit from vroom. But, if the argument only exists in vroom, like col_select, those docs should inherit from vroom?

Yeah that is certainly the way I'm currently thinking about it.

PR #1436, opened 2022-09-13
send `spec_csv()` to vroom

### LINKED ISSUE #1387: Name repair different for read_csv() and spec_csv()
When trying to create column specifications of a file that contains duplicated variable names, spec_csv() renames the variables differently (e.g "error_1") than read_csv() ("error...1").

Let test.csv be a simple CSV with duplicated variables "error" and one observation:

```sh
echo "a,error,b,error,c,error\n1,string1,2,string2,3,string3" >> test.csv
```

```r
readr::spec_csv("test.csv")
```
```
cols(
  a = col_double(),
  error = col_character(),
  b = col_double(),
  error_1 = col_character(),
  c = col_double(),
  error_2 = col_character()
)

Warning message:
Duplicated column names deduplicated: 'error' => 'error_1' [4], 'error' => 'error_2' [6] 
```
```r
readr::read_csv("test.csv")
```

```
New names:                                                                                                                                                                                                  
* error -> error...2
* error -> error...4
* error -> error...6
Rows: 1 Columns: 6
── Column specification ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Delimiter: ","
chr (3): error...2, error...4, error...6
dbl (3): a, b, c

ℹ Use `spec()` to retrieve the full column specification for this data.
ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
# A tibble: 1 × 6
      a error...2     b error...4     c error...6
  <dbl> <chr>     <dbl> <chr>     <dbl> <chr>    
1     1 string1       2 string2       3 string3
```

sbearrows:
Fixes #1387 

Previously, `spec_*()` was being routed to readr ed1 even when using ed2. This meant that duplicate name repair was not consistent between `spec_*()` and the equivalent `read_*()` function.

### DIFF
--- a/man/spec_delim.Rd
+++ b/man/spec_delim.Rd
@@ -37,7 +37,7 @@ spec_delim(
 spec_csv(
   file,
   col_names = TRUE,
-  col_types = list(),
+  col_types = NULL,
   col_select = NULL,
   id = NULL,
   locale = default_locale(),
@@ -47,12 +47,11 @@ spec_csv(
   comment = "",
   trim_ws = TRUE,
   skip = 0,
-  n_max = 0,
   guess_max = 1000,
   name_repair = "unique",
   num_threads = readr_threads(),
   progress = show_progress(),
-  show_col_types = should_show_types(),
+  show_col_types = FALSE,
   skip_empty_rows = TRUE,
   lazy = should_read_lazy()
 )
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -499,7 +499,59 @@ spec_delim <- generate_spec_fun(read_delim)
 
 #' @rdname spec_delim
 #' @export
-spec_csv <- generate_spec_fun(read_csv)
+spec_csv <- function(file,
+                     col_names = TRUE,
+                     col_types = NULL,
+                     col_select = NULL,
+                     id = NULL,
+                     locale = default_locale(),
+                     na = c("", "NA"),
+                     quoted_na = TRUE,
+                     quote = "\"",
+                     comment = "",
+                     trim_ws = TRUE,
+                     skip = 0,
+                     guess_max = 1000,
+                     name_repair = "unique",
+                     num_threads = readr_threads(),
+                     progress = show_progress(),
+                     show_col_types = FALSE,
+                     skip_empty_rows = TRUE,
+                     lazy = should_read_lazy()) {
+  if (edition_first()) {
+    tokenizer <- tokenizer_csv(
+      na = na, quoted_na = quoted_na, quote = quote,
+      comment = comment, trim_ws = trim_ws, skip_empty_rows = skip_empty_rows
+    )
+    return(spec(read_delimited(file, tokenizer,
+      col_names = col_names, col_types = col_types,
+      locale = locale, skip = skip, skip_empty_rows = skip_empty_rows,
+      comment = comment, n_max = 0, guess_max = guess_max, progress = progress,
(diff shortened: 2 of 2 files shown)


sbearrows:
I've only implemented this for `spec_csv()` so we can iron things out first.

man/spec_delim.Rd
```diff
@@ -47,12 +47,10 @@ spec_csv(
   comment = "",
   trim_ws = TRUE,
   skip = 0,
-  n_max = 0,
   guess_max = 1000,
   name_repair = "unique",
   num_threads = readr_threads(),
   progress = show_progress(),
-  show_col_types = should_show_types(),
   skip_empty_rows = TRUE,
```
sbearrows, line 54: Since `n_max` will always `= guess_max` I've removed it from the function arguments. I also removed `show_col_types()` because otherwise the column specification would print twice.

jennybc:
Yes, this is a faithful implementation of the proposal in https://github.com/tidyverse/readr/issues/1387#issuecomment-1232245963:

```
spec_csv_vroom <- function(..., guess_max = 1000) {
  tmp <- readr::read_csv(..., n_max = guess_max, guess_max = guess_max)
  spec(tmp)
}
```

It looks like the right first move.

man/spec_delim.Rd
jennybc: When I first looked at `spec()` because of this issue, I was sort of surprised at how many arguments there are.

So, yes, I would definitely take a hard look at whether all of them are actually being used / make sense. OTOH, especially, since readr 2e is going to basically parse the file to get the spec, it does makes sense that almost all of the usual arguments are here.

jennybc (commented):
Yes, this is a faithful implementation of the proposal in https://github.com/tidyverse/readr/issues/1387#issuecomment-1232245963:

```
spec_csv_vroom <- function(..., guess_max = 1000) {
  tmp <- readr::read_csv(..., n_max = guess_max, guess_max = guess_max)
  spec(tmp)
}
```

It looks like the right first move.

sbearrows:
Carrying a conversation about guessing behaviors in `read_*()` vs `spec_*()` over from #1431 

One consequence of doing the proposal from #1387 is the following:

Since `guess_max` doesn't guess column types using rows it will never read, if we complete this PR as intended then `n_max = guess_max` and guessing for `spec_*()` will never be done via selecting rows "interspersed throughout the file". For example, if some data has 2000 rows and `guess_max = 1000` then `n_max = 1000` and it will guess using the first 1000 rows in the data. It'll always use all of the data it is given to guess. Here is an example using the challenge.csv where the column y is filled with NA until row 1001, after which it becomes filled with dates.

```r
# after we complete #1436
spec_csv(readr_example("challenge.csv"), guess_max = 1000) # n_max = 1000
#> cols(
#>   x = col_double(),
#>   y = col_logical()
#> )
# increasing guess_max does change the spec
spec_csv(readr_example("challenge.csv"), guess_max = 1001) # n_max = 1001
#> cols(
#>   x = col_double(),
#>   y = col_date(format = "")
#> )

# compared to how guessing works in read_csv
# since it's dispersed throughout the file, it guesses correctly
spec(read_csv(readr_example("challenge.csv"), guess_max = 1000, show_col_types = FALSE)) # n_max = all the data
#> cols(
#>   x = col_double(),
#>   y = col_date(format = "")
#> )
```

Not necessarily a non-starter, but food for thought.

PR #1439, opened 2022-09-30
Soft-deprecation warning is no longer signalled

lionel-:
Because the deprecated feature (from `vroom:::standardise_path()`) is not called directly, and we no longer warn in that case for soft deprecations in the upcoming version of lifecycle.

It seems like the test is expecting another warning to happen but it doesn't. I'm not sure though.


jennybc:
I have to release readr today and I've just put a skip in. I'll remove that and apply this fix after readr is on CRAN.

PR #1463, opened 2023-01-22
`write_excel_csv()` and `write_excel_csv2()` put datetimes in a format…

### LINKED ISSUE #1202: feature request: recognizable format datetimes `write_excel_csv`
As described in #553 and #948 datetimes written by `write_excel_csv` now have the format `2018/01/01 00:00:00`. The pity is that `readr`'s read functions do not recognize this format. Can it be changed to a dashed format, because afaics Excel also accepts that?

```r
str(readr::parse_guess("2018/01/01 00:00:00"))
#>  chr "2018/01/01 00:00:00"
str(readr::parse_guess("2018-01-01 00:00:00"))
#>  POSIXct[1:1], format: "2018-01-01"

<sup>Created on 2021-04-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

```

keesdeschepper:
… that is readable again by `readr` on roundtrips.  Best `strftime()` format for excel csv datetimes according to https://stackoverflow.com/questions/804118/best-timestamp-format-for-csv-excel#comment57807590_804137 Fixes #1202.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # readr (development version)
 
+`write_excel_csv()` and `write_excel_csv()` put datetimes in a format that is
+readable again by `readr` on roundtrips (#1202).
+
 # readr 2.1.3
 
 * Help files below `man/` have been re-generated, so that they give rise to valid HTML5. (This is the impetus for this release, to keep the package safely on CRAN.)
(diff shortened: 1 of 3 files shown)


jennybc:
Just an FYI on why I didn't move on this, I see the point but it would also be a breaking change in some sense, so requires some deliberation.

keesdeschepper:
hi @jennybc, how is the deliberation going?

uaesooraj08-cpu (approved):

PR #1548, opened 2024-07-28
fix html

spaette:
this is a copy/paste from googlesheets4

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -46,7 +46,7 @@ pak::pak("tidyverse/readr")
 
 ## Cheatsheet
 
-<a href="https://raw.githubusercontent.com/rstudio/cheatsheets/main/data-import.pdf"><img src="https://github.com/rstudio/cheatsheets/raw/main/pngs/thumbnails/data-import-cheatsheet-thumbs.png" height="252" alt="thumbnail of tidyverse data import cheatsheet"//></a>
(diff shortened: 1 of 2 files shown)


jennybc:
This is not necessarily critique but I'm wondering: are you detecting a problem with the current site? There's no problem with the naked eye for me.

spaette:
> current site

as opened in Chromium and Firefox

likely due to `` element's extra `/` at the end

```
<img src="https://github.com/rstudio/cheatsheets/raw/main/pngs/thumbnails/data-import-cheatsheet-
thumbs.png" height="252" alt="thumbnail of tidyverse data import cheatsheet"//>
```

> copy/paste from googlesheets4

this does two things regarding the user interface

- renders cheatsheets/data-import.pdf
- eliminates the redirect occurring for the image file

a download icon accompanies the .pdf rendering whereas currently the file is downloaded to a default directory

jennybc:
OK I'm happy to update. The extra slash is clearly a mistake that I guess some browsers cope with.

The official workflow is to edit `README.Rmd` as per:

https://github.com/tidyverse/readr/blob/b4cb4f7270252259c30fa4dbaaee32ba9ccea0f1/README.Rmd#L5

then call `devtools::build_readme()`. If you don't update the PR, I will next time I come through here.

spaette:
likely I'll update tomorrow

jennybc:
Thanks!

PR #1618, opened 2026-02-22
Mention clipboard() in read_delim() file param docs

### LINKED ISSUE #1614: `clipboard()` is not mentioned as possible value for, e.g., read_csv(file =)
In c2a334f192ff0160a2e5e9afc0982ad1bcf543af, I switched to inheriting more docs (specifically for `file`) from `vroom::vroom()` as opposed to from `readr::datasource()`. This is a net positive as it better reflects actual reality!

But it is true that readr offers `clipboard()` for posssible use with `file` and that's no longer mentioned (vroom does not export such a function). I can live with this omission for now and maybe forever.

This is a note-to-self to revisit this when I assimilate vroom into readr, which changes the landscape.

chrislongros:
Closes #1614.

`read_delim()` inherits its `file` param docs from `vroom::vroom()`, which doesn't mention `clipboard()`. Other read functions (`read_table()`, `read_fwf()`, etc.) inherit from `datasource()` and already document it. This adds an explicit `@param file` to `read_delim()` that includes the `clipboard()` mention.

### DIFF
--- a/R/read_delim.R
+++ b/R/read_delim.R
@@ -10,6 +10,22 @@ NULL
 #' decimal point. This format is common in some European countries.
 #' @inheritParams tokenizer_delim
 #' @inheritParams vroom::vroom
+#' @param file Either a path to a file, a connection, or literal data
+#'   (either a single string or a raw vector). `file` can also be a
+#'   character vector containing multiple filepaths or a list containing
+#'   multiple connections.
+#'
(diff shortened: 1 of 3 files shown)


jennybc:
Thanks but I'm deliberately inheriting docs from vroom and will soon bring vroom code into readr. That's going to precipitate alot of docs work and I'll deal with #1614 then.

PR #1624, opened 2026-05-18
Add date_order locale option and flexible date separator parsing

hidekoji:
## Summary

Adds a `date_order` argument to `locale()` and makes date / date-time auto-detection more forgiving, so year-last dates (e.g. `10/02/2024`) can be parsed as dates instead of guessed as character.

* **`locale(date_order =)`** — new optional argument accepting an explicit component order: `"ymd"`, `"mdy"`, `"dmy"`, etc., optionally with a time suffix (`"mdy_hms"`, `"dmy_hm"`, `"ymd_h"`). `NULL` (default) keeps current behaviour. Validated in R with a clear error message.
* **`DateTimeParser::parseDateOrder()`** — parses a value against an explicit order, including an optional `T`/space-separated time part.
* **`DateTimeParser::parseYearLastHeuristic()`** — recognises unambiguous `D/M/YYYY` vs `M/D/YYYY` (`part > 12` disambiguates; defaults to MDY when ambiguous). Used as an auto-detection fallback in both `isDate()` (guesser) and `CollectorDate::setValue()` (collector) so the two agree.
* **Flexible separators** — date parsing accepts any non-alphanumeric separator between components (`2024.10.02`, `2024/10/02`, …).
* **Collectors** — `CollectorDate` / `CollectorDateTime` hold a `LocaleInfo*` and dispatch through `parseDateOrder()` when `date_order` is set with no explicit format.

## Scope / known limitation

These changes affect readr's own C++ engine, which powers `parse_date()`, `parse_datetime()`, `guess_parser()`, and the **first edition** of `read_csv()`.

The **second edition** of `read_csv()` (default since readr 2.0) delegates parsing to the **vroom** package, which does not yet know about `date_order`. Full end-to-end `read_csv()` support in the default edition therefore depends on the companion vroom change (tidyverse/vroom#623). End-to-end `read_csv()` tests are intentionally omitted from this PR until that lands.

## Test plan

- [x] `tests/testthat/test-locale.R` — `locale()` accepts valid `date_order`, rejects invalid values, defaults to `NULL`
- [x] `tests/testthat/test-parsing-datetime.R` — `guess_parser()` detects MDY/DMY dates and datetimes with explicit `date_order`
- [x] `guess_parser()` auto-detects unambiguous DMY year-last dates; ambiguous year-last defaults to MDY
- [x] `parse_date()` / `parse_datetime()` honour `locale(date_order =)`
- [x] existing YMD / ISO-8601 behaviour unchanged (backward compatibility)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # readr (development version)
 
+* `locale()` gains a `date_order` argument to control the component order used
+  when parsing dates and date-times (e.g. `"mdy"`, `"dmy"`, `"ymd_hms"`). This
+  makes it possible to read year-last formats such as `10/02/2024` that the
+  automatic type guesser would otherwise treat as character.
+
+* Date and date-time auto-detection now accepts any non-alphanumeric separator
+  between components (e.g. `2024.10.02`, `2024/10/02`), and falls back to a
+  year-last heuristic so unambiguous `D/M/YYYY` values are recognised as dates.
+
 # readr 2.2.0
 
 This release advances many deprecations.
--- a/R/locale.R
+++ b/R/locale.R
@@ -32,6 +32,10 @@
 #' @param asciify Should diacritics be stripped from date names and converted to
 #'   ASCII? This is useful if you're dealing with ASCII data where the correct
 #'   spellings have been lost. Requires the \pkg{stringi} package.
+#' @param date_order Order of date components for auto-detection. One of
+#'   `"ymd"`, `"ydm"`, `"mdy"`, `"myd"`, `"dmy"`, `"dym"`, or those combined
+#'   with a time suffix: `"_hms"`, `"_hm"`, or `"_h"` (e.g. `"mdy_hms"`).
+#'   Use `NULL` (default) for automatic detection.
 #' @export
 #' @examples
 #' locale()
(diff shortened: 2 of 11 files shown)


hidekoji:
Added a commit extending the year-last date heuristic to **2-digit years** (e.g. `5/29/26`), plus a **year-last datetime** heuristic:

- `consumeYearFlexible()` consumes a 2- or 4-digit year and applies the standard `%y` pivot (00–68 → 2000s, 69–99 → 1900s), rejecting implausible 3-digit values; the year-last date heuristic and the `date_order` year component route through it.
- `parseYearLastHeuristicDateTime()` recognizes `M/D/Y` / `D/M/Y` date + `HH[:MM[:SS]]` (optional tz); wired into the guesser and the materialization path so MDY/DMY datetimes are detected.
- New tests: 2-digit MDY/DMY/ambiguous dates, the `%y` pivot, invalid/3-digit rejection, 2- and 4-digit MDY datetimes, and explicit `date_order` with 2-digit years. Existing date/datetime tests unchanged (incl. "%Y requires 4 digits" — explicit `%Y`/`%y` format paths are untouched). `DESCRIPTION` not bumped.

Issue #1247, opened 2021-07-27
write_file cannot write files with ".zip" extension


MilesMcBain:
`write_file` cannot write .zip files. The issue seems to be in this switch statement in `standardise_path`:  https://github.com/tidyverse/readr/blob/31236cf59a32173e47954e4be45ad44c6504695f/R/source.R#L161

Where the other zip functions create connections, `readr::zipfile` attempts to decompress the file file, which doesn't exist because we're trying to write it.

``` r
readr::write_file(charToRaw(letters), "./test.zip")
#> Error in utils::unzip(path, list = TRUE): zip file './test.zip' cannot be opened
```

<sup>Created on 2021-07-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

jimhester:
Base R does not support writing to zip files from a connection. e.g. there is no equivalent to `unz()`.

However you can now use the [archive package](https://archive.r-lib.org/) to do this.

```r
readr::write_file(paste0(letters, collapse = ""), archive::archive_write("letters.zip", "letters"))
```

MilesMcBain:
Ah sorry if I was unclear. I don't want to zip data. I have the raw data representing a zip file in the body of a http request. I can't write that to a .zip file on disk. 

My work around was to write it with a different extension, then change the extension to zip after.

minemR:
My work around was to use:
```
readr:::write_file_raw_(x, connection)
```
I guess that is quite bad, because the unexported function could change or be removed in next release...

jimhester:
Perhaps just use `base::writeBin()` if you want to write bytes to disk and not do any transformations of the data.

cortinah:
hello and thanks all, coming back to an old issue.  I used readr::write_csv(dataframe, "filename.csv.zip") today and it works fine.  Writes a .zip file.  Looking at the vroom code, I can see that it supports 4 compression schemes, including .zip.   So i think this issue can be closed.   I can submit a PR to update readr documentation since it states that only 3 compression schemes are supported, but I think it is 4 (.bz2. .gz, .xz, and .zip).  thanks!

PR #1625, opened 2026-06-01
Update write_ documentation to indicate .zip support

cortinah:
vroom supports writing .zip files and so does readr::write_csv() but documentation doesn't state that.


cortinah:
closes #1247

PR #1638, opened 2026-06-07
docs: fix tokenize() example param, add @return to tokenize() and count_fields()

LeonidasZhak:
## Changes

- **Fix `tokenize()` example**: Change `n = 2` to `n_max = 2` in the example. The old code worked via R's partial argument matching, but used the wrong parameter name in documentation.

- **Add `@return` to `tokenize()`**: Documents that it returns a list of character vectors, one per line.

- **Add `@return` to `count_fields()`**: Documents that it returns an integer vector with one element per line.

## Validation

- Tests pass: `devtools::test(filter = "parsing-character|tokenize|count-fields")` — 40 pass, 0 fail
- Full test suite: 615 pass, 30 pre-existing failures (related to vroom `problems` attribute propagation, unrelated to this change)

### DIFF
--- a/R/count_fields.R
+++ b/R/count_fields.R
@@ -8,6 +8,8 @@
 #'   up into fields, e.g., [tokenizer_csv()],
 #'   [tokenizer_fwf()]
 #' @param n_max Optionally, maximum number of rows to count fields for.
+#' @return An integer vector with one element per line, giving the number of
+#'   fields in each line.
 #' @export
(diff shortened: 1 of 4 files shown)


LeonidasZhak:
Withdrawing this small automated PR while I consolidate an oversized batch of contributions and reduce maintainer review burden. Sorry for the noise, and thank you for maintaining the project.

